PR #152, opened 2013-11-20
New: stacked barplot

jonathanrocher:
This is bringing back to life a PR by Travis Vaught to add a new type of plot in Chaco: a stacked barplot, leading plots of this type:

screen shot 2013-11-20 at 12 53 55 pm


corranwebster:
And at least a "smoke test"-style unit test!

PR #162, opened 2013-12-13
RectangleSelectionTool

timdiller:
This PR adds a RectangleSelectionTool based on the BetterSelectingZoom tool (thanks @tonysyu for the initial implementation). As such, it includes an `overlay()` method and subclasses `AbstractOverlay` as well as `BaseTool` so that it can be appended both to the tools and overlays lists.

Included is a demo `chaco/examples/demos/basic/rectangle_selection.py` which uses the rectangle tool to select a region in one plot to display zoomed in the other plot.

image


tmreay:
I'm currently working on building a tool to crop regions of an image, and this overlay has proved very useful.

We don't necessarily need to merge it because of my use case, but it should probably be documented that someone else has found this useful since it was originally written.

notmatthancock:
Is this PR superseded by #482 ?

corranwebster:
They are different enough that this probably shouldn't be closed - this looks like it is designed for selecting rectangular regions of image plots, not scatter plots.

timdiller:
It sounds like this tool could use a name change to highlight the difference from `RectangularSelection`.
I propose `RectangleRegionSelection`, with appropriate updates to docstrings to make it clear what this is good for.  It might be worth adding a section in the user docs discussing the differences, similar to `TableEditor` and `TabularEditor` in `traitsui`.  @corranwebster if that seems like a good idea, I'm willing to take it on.

PR #194, opened 2014-05-07
Add Dash/Zeal docset make target

johntyree:
Just like enthought/traits#180, this PR introduces a make target for a Dash/Zeal type docset.

It assumes you have doc2dash installed and keeps all output under $(BUILDDIR).

The docset_html target lacks the left sidebar to make viewing in Dash/Zeal nicer. Unfortunately, sphinx does not support the notion of configuration files named anything other than conf.py and so we must make a directory docset_resources in which to hold the modified configuration. If there is a way to tell sphinx-build to override the html_theme_options dictionary then this directory is unneeded.


jvkersch:
@johntyree If you're still around, I took the liberty of (messily) merging master in your branch and fixing the one issue Chris suggested. This PR has actually been quite useful to me; I built the Dash docset a while ago and have been using it ever since (unfortunately without actually pinging you on this PR :( )

jvkersch:
@cfarrow I addressed your comment from 2 years ago in this commit: https://github.com/enthought/chaco/pull/194/commits/c8c3f616a119b73f9c4cbe51b2b57c2bc8957fb0. Does this look good to you?

cfarrow:
LGTM

PR #241, opened 2014-12-12
Add default position and tests for BetterZoom

### LINKED ISSUE #240: BetterZoom position attribute not initialized
The `position` attribute used by the `BetterZoom` tool only gets set by mouse moves.  Interactions which cause zooming prior to the mouse entering the component the tool is attached to will throw an exception when they try to access the attribute.

A sensible default should probably be provided for the position (eg. the centre of the component).

corranwebster:
Fixes #240.

The problem occurred when the tool was added with keyboard focus on the component and the mouse in the region where the tool was working.  A keyboard interaction with the tool would then give a traceback because `position` was not defined.

### DIFF
--- a/chaco/tools/better_zoom.py
+++ b/chaco/tools/better_zoom.py
@@ -1,4 +1,13 @@
-import numpy
+# Copyright (c) 2005-2014, Enthought, Inc.
+# All rights reserved.
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only
+# under the conditions described in the aforementioned license.  The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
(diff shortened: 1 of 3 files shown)


chaco/tools/better_zoom.py
```diff
@@ -287,13 +298,11 @@ def _zoom_limit_reached(self, factor, xy_axis):
         """
 
         if xy_axis == 'x':
-            if factor <= self.x_max_zoom_factor and factor >= self.x_min_zoom_factor:
-                return False
-            return True
+            return (factor > self.x_max_zoom_factor or
```
mdickinson: `return not self.x_min_zoom_factor <= factor <= self.x_max_zoom_factor` ?

mdickinson:
LGTM.

PR #260, opened 2014-12-23
Fix ticks inf

Snegovikufa:
Update implementation of https://github.com/enthought/chaco/pull/126


chaco/ticks.py
```diff
@@ -261,6 +272,11 @@ def heckbert_interval(data_low, data_high, numticks=8):
     return graphmin, graphmax, d
 
 
+def _isOutOfRange(value):
+    """ determines if value is float and equals to infinity """
+    return value is float and isinf(value)
```
Snegovikufa: > You probably wanted isinstance(value, (float, numpy.floating)).

Yeah, my bad.

chaco/ticks.py
```diff
@@ -205,7 +209,14 @@ def auto_ticks ( data_low, data_high, bound_low, bound_high, tick_interval,
 
     if upper > end:
         end += tick_interval
-    ticks = arange( start, end + (tick_interval / 2.0), tick_interval )
+
+    if isinf(start) or isinf(end + (tick_interval / 2.0)):
```
Snegovikufa: You're wrong

chaco/tests/test_ticks.py
```diff
@@ -0,0 +1,60 @@
...
+        else:
+            x *= increment
+
+    return prev_x
+
+
+class TestTicks(unittest.TestCase):
+
+    def testAutoTicksOnInfinity(self):
+        inf = 1e308 * 1e308
```
Snegovikufa: Nice suggestation! :+1:

chaco/tests/test_ticks.py
```diff
@@ -0,0 +1,60 @@
+from math import isinf
+import unittest
+from chaco.ticks import auto_ticks
+
+__author__ = 'Rustam Safin'
+
+
+def get_near_inf_value():
```
Snegovikufa: Nice :+1:

chaco/ticks.py
johntyree: Hmmm... I agree with you. Regardless of `tick_interval`, `start` will be a `float` because that's what `floor` will always return. Not sure what that other guy was thinking.

johntyree: Actually, `isinf` doesn't break on ints anyway...  I have no idea what that was about.

PR #261, opened 2014-12-23
Python 3 support for 4.6.x branch

corranwebster:
This is a PR to merge the Python 3 support into the backwards compatible 4.6 branch.  Before this can be merged into master there should be similar merges of the Python 3 support into Pyface, TraitsUI, and Enable master.

This was made as a request from #259, but I'm personally not interested in pushing this further right now.  If someone wants to pick it up and run with it, feel free.


corranwebster:
OK, we are green on tests.  I'd like to merge this sooner rather than later and worry about any bugfixes later.

@itziakos @cfarrow Any reason not to merge?

chaco/base.py
```diff
@@ -84,7 +86,8 @@ def bin_search(values, value, ascending):
     Returns -1 if `value` is beyond the minimum or maximum of `values`.
 
     """
-    if ascending > 0:
+    ascending = ascending > 0
+    if ascending:
```
itziakos: (minor) I think this change is uncessary

chaco/color_mapper.py
```diff
@@ -304,7 +305,7 @@ def _get_color_bands(self):
         if self.color_depth is 'rgba':
             luts.append(self._alpha_lut)
 
-        result = zip(*luts)
+        result = list(zip(*luts))
```
itziakos: This also look wierd. It will also place a performance penalty on python 2.7./ Why not use `itertools.izip` in the first place

chaco/scales/formatters.py
```diff
@@ -236,7 +239,7 @@ class IntegerFormatter(BasicFormatter):
     def format(self, ticks, numlabels=None, char_width=None, fill_ratio=0.3):
         """ Formats integer tick labels.
         """
-        return map(str, map(int, ticks))
+        return list(map(str, map(int, ticks)))
```
itziakos: I would suggest to use `itertools.imap` so that python 2.7 does not get a performance penalty for creating and destroying multiple list objects

chaco/tests/test_colormapped_scatterplot.py
```diff
@@ -1,12 +1,18 @@
...
 
 from numpy import alltrue, arange
+
 from enable.compiled_path import CompiledPath
+from traits.testing.unittest_tools import unittest
 
 # Chaco imports
-from chaco.api import (ArrayDataSource, ColormappedScatterPlot, DataRange1D,
-                       LinearMapper, PlotGraphicsContext, jet)
+from chaco.array_data_source import ArrayDataSource
```
itziakos: Is there a reason for this change?

itziakos:
@corranwebster, I had a quick look. It looks to me that some of the code commited was the result of already applying the `py2to3` script on the files. This was unnecessary because setuptools will make sure that py2to3 is applied when building for python 3 and probably bad because the result of py2to3 is not always the best choice for python 2.7 (e.g. wrapping zip into a list)

chaco/base.py
corranwebster: Yes - the need for it got removed in a subsequent change.

corranwebster:
There was a period over a year ago where I was trying to get this to run from a single codebase without 2to3, which is probably where some of that crept in.  It was also around the time I was learning to do Python 3 transltions, so I think I did some stuff that wasn't needed.

Having done a little work in the Python 3 world now, it would be nice if we could get it so we didn't need 2to3 in the setup.py; but not for this PR, I think.

itziakos:
My problems with this PR are:

1. py2to3 results have been committed.
2. A number of unnecessary changes exist (e.g unraveling imports).
3. The `from __future__` imports are added even when they are not needed.

I do not think that it should be merged. Looking at the changes one should be able to recreate the PR without much effort (most of the changes should happen automatically by py2to3 fixers)

corranwebster:
Closing in favour of #301

PR #263, opened 2014-12-30
Speed-ups for chaco.base

corranwebster:
Replace `bin_search` with a `numpy.searchsorted`-based algorithm. This gives a 5x speedup, but keeps the results the same (unit tests should all pass).

Also use `column_stack` instead of `zip` in `arg_find_runs`.  This is slightly slower in typical use cases, but up to 10x faster in bad cases (large arrays with many separate runs).  As part of this, added tests for `arg_find_runs` and the `flat` argument, since this is what is actually used in practice in a lot of chaco code.

### DIFF
--- a/chaco/base.py
+++ b/chaco/base.py
@@ -61,12 +61,28 @@ def n_gon(center, r, nsides, rot_degrees=0):
 
 
 def bin_search(values, value, ascending):
-    """
-    Performs a binary search of a sorted array looking for a specified value.
+    """ Performs a binary search of a sorted array for a specified value.
+
+    Parameters
+    ----------
+
+    values : array
+        The values being searched.
+
+    value : float
+        The value being searched for.
+
+    ascending : -1 or 1
+        This value should be 1 if the values array is ascending, or -1 if
+        the values array is descending.
+
+    Returns
+    -------
 
     Returns the lowest position where the value can be found or where the
     array value is the last value less (greater) than the desired value.
     Returns -1 if `value` is beyond the minimum or maximum of `values`.
+
     """
     if ascending > 0:
         if (value < values[0]) or (value > values[-1]):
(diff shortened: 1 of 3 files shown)


chaco/base.py
```diff
@@ -60,34 +60,25 @@ def n_gon(center, r, nsides, rot_degrees=0):
...
 
-# Ripped from Chaco 1.0's plot_base.py
 def bin_search(values, value, ascending):
     """
     Performs a binary search of a sorted array looking for a specified value.
 
     Returns the lowest position where the value can be found or where the
     array value is the last value less (greater) than the desired value.
-    Returns -1 if *value* is beyond the minimum or maximum of *values*.
+    Returns -1 if `value` is beyond the minimum or maximum of `values`.
```
pberkes: Could you please add some info about the value of `ascending`?

chaco/tests/base_utils_test_case.py
```diff
@@ -133,6 +134,49 @@ def test_find_runs_descending(self):
         assert_equal(find_runs(x, order='descending') , \
                             [[30], [41,40,39,38,37], [12]])
 
+    def test_find_runs_flat(self):
+        x = array([0,0,0,1,1,1,1,0,0,0,0])
+        assert_equal(find_runs(x, order='flat') , \
+                            [[0,0,0], [1,1,1,1], [0,0,0,0]])
```
pberkes: Unnecessary `\` and strange indentation.

pberkes: Same in tests below

pberkes:
Looks good, only minor comments

chaco/base.py
corranwebster: Done.

corranwebster:
Comments addressed, tests passing.  Merging.

PR #269, opened 2015-04-09
Automatic, PEP386-compliant version number.

pberkes:
Based on John Wiggin's work for traits, codetools, etc.


chaco/__init__.py
```diff
@@ -3,7 +3,7 @@
 """ Two-dimensional plotting application toolkit.
     Part of the Chaco project of the Enthought Tool Suite.
 """
-__version__ = '4.6.0-dev'
+from ._version import full_version as __version__
```
itziakos: flake8 complains about `__version__` imported but not used. Should we include an `__all__` definition?

setup.py
```diff
@@ -1,77 +1,174 @@
...
+    with open(filename, "wt") as fp:
+        fp.write(template.format(version=VERSION,
+                                 full_version=fullversion,
+                                 git_revision=git_rev,
+                                 is_released=IS_RELEASED))
+
+
+if __name__ == "__main__":
+    write_version_py()
+    from chaco import __requires__, __version__
```
itziakos: This looks like an old style relative import. I think it will cause problems with python 3 in the future.

rkern: How so? setup.py isn't in a package.

itziakos: Yup, you are right.  My mistake. Please ignore my comment.

chaco/__init__.py
pberkes: I vote for a mere `# noqa`

itziakos:
:+1:  to me, just minor comment.

itziakos:
The travis-ci failure however, looks fun

chaco/__init__.py
itziakos: As long as flake8 does not complain I am ok.

dev_requirements.txt
```diff
@@ -6,3 +6,4 @@ coverage
 -e git+http://github.com/enthought/traits.git#egg=traits
 -e git+http://github.com/enthought/traitsui.git#egg=traitsui
 -e git+http://github.com/enthought/enable.git#egg=enable
+-e git+http://github.com/nucleic/kiwi.git#egg=kiwisolver
```
itziakos: I would suggest to use the offcial pypi egg for kiwisolver

itziakos:
updated converalls.io config to only use the github status api

pberkes:
Please don't merge yet, pending better git tag matching string.

pberkes:
As far as I'm concerned, this is ready for a final review.

setup.py
```diff
@@ -1,77 +1,181 @@
...
+        env['LC_ALL'] = 'C'
+        out = subprocess.Popen(
+            cmd, stdout=subprocess.PIPE, env=env,
+        ).communicate()[0]
+        return out
+
+    try:
+        out = _minimal_ext_cmd(
+            ['git', 'describe', '--match', '[0-9]*.[0-9]*.[0-9]*', '--tags'])
+    except OSError:
```
jonathanrocher: Could you add a comment with what you are expecting here?

pberkes: Good point, I will

setup.py
pberkes: Comment ignored.
;-)

setup.py
pberkes: Done

pberkes:
@jonathanrocher Any other comment?

jonathanrocher:
Sorry. LGTM. Merging.

mdickinson:
For some reason, since this PR was merged, a `python setup.py develop` hangs when trying to install Chaco's dependencies, entering some kind of infinite memory-hungry loop.  I've reproduced this both with and without Canopy, on Linux and on OS X.  Can anyone else reproduce?

PR #270, opened 2015-04-13
Update old urls

kjordahl:
A link to the enable docs was broken, updated to a current working link.

Also updates the `url` in `setup.py` to this GitHub repo. Other ETS projects are not consistent about this; for example [traits has an old link](https://github.com/enthought/traits/blob/master/setup.py#L118) while [traitsui links to GitHub](https://github.com/enthought/traitsui/blob/master/setup.py#L35).


docs/source/user_manual/containers.rst
```diff
@@ -563,5 +563,5 @@ but uses instead
 its own reference to the window, :attr:`self.window`.
 
 See also the `documentation of the enable library
-<http://github.enthought.com/enable/enable_concepts.html>`_, which gives
+<http://docs.enthought.com/enable/enable_concepts.html>`_, which gives
```
jonathanrocher: +1

setup.py
```diff
@@ -41,7 +41,7 @@
     author_email = 'pwang@enthought.com',
     maintainer = 'ETS Developers',
     maintainer_email = 'enthought-dev@enthought.com',
-    url = 'http://code.enthought.com/projects/chaco',
+    url = 'https://github.com/enthought/chaco'
```
jonathanrocher: For this one I would keep it the way it is since my understanding is that code.enthought.com is the home page of the project and github is where we host the documentation. I know code.e.com is old and bad looking, but I don't think we have officially decided to throw it out. @dpinte @rkern an opinion?

kjordahl: Apparently traitsui has had the url pointing to GitHub [since 2011](https://github.com/enthought/traitsui/commit/ab23cf1f4f739d76c3eff3f2ea3a074cdedeadbb).

We should at least be consistent and make a deliberate choice about it.

jonathanrocher: Absolutely agree.

corranwebster:
See this document for the correct values for links (its a fair way down the page in the release process notes): https://github.com/enthought/documentation/blob/master/source/guidelines/ets_maintenance.rst

In short, `url` should point to the appropriate docs.enthought.com page, and `download_url` should point to the github project.

setup.py
```diff
@@ -41,7 +41,7 @@
     author_email = 'pwang@enthought.com',
```
jwiggins: This should probably be info@enthought.com, no?

jonathanrocher: Indeed. That email address is not valid anymore.

jonathanrocher:
@kjordahl ok with you if I resolve the conflict with master and merge this in?

kjordahl:
@jonathanrocher Please do!

corranwebster:
Update to containerized builds, tests pass, lets' merge this!

PR #275, opened 2015-08-27
More colormaps & DiscreteColormap class

corranwebster:
Add extra colormaps from various sources:

* the colorbrewer quantitative colormaps (extended from discrete to continuous by linear interpolation, for better or worse, same as matplotlib)
* the 'terrain' and 'seismic' colormaps from matplotlib (and possibly other places)
* the 'wistia' colorblind-friendly heatmap from https://github.com/wistia/heatmap-palette (MIT licensed)
* the 'magma', 'inferno', 'plasma' and 'viridis' colormaps recently introduced to Matplotlib


rkern:
We explicitly removed the quantitative Colorbrewer colormaps. They are not appropriate interpolated.

corranwebster:
Nevertheless, they are available in matplotlib, and were useful in a consulting project.  There is some value in having some parity with matplotlib's capabilities.

rkern:
I strongly disagree. Parity with bad decisions does us no good.

corranwebster:
Looks like we are having CI build errors related to PIL/Pillow stuff.

kjordahl:
To be clear, the issue is over the *qualitative* (not quantitative) colormaps here, correct?

I agree that they are flawed, but also that they are useful. Was there previous discussion of this for chaco?

rkern:
Yes, the qualitative (categorial) colormaps that have N discrete, unordered colors.

If you want to use them, implement a discrete colormapper. Then I will have no objection to exposing them as such. Exposing them as continuous interpolated colormaps encourages misuse.

The discussion was back in late 2007.

rkern:
If you make these colormap functions have a fixed range from `-0.5` to `N-0.5` and with constant color from `i-0.5` to `i+0.5` for each color `i`, then that would be fine. Then these could be used well with integer data in `range(0, N)`. Users will have to map their qualitative data to those integers, but that's pretty much par for the course. It would be ideal to keep them in a separate list from the other default continuous colormaps.

corranwebster:
I'm currently sketching out a simple `DiscreteColorMapper` with similar semantics (essentially holds an array of colors, ignores the data range, expects integers from 0 to N-1).  It's not terribly complex, and with fancy indexing I think will be reasonably fast.

I'm tempted to truncate, rather than round if given floats, with the thought that if a user is worried about floating point rounding issues then they should deal with that themselves before giving the mapper the data.

rkern:
LGTM! I think all of the failures are PIL/Pillow dependency-related. I'm not sure if we have a fix for that or not.

PR #276, opened 2015-08-28
Use travis ci containers

itziakos:


itziakos:
PR fails mainly because of issues with the new Pillow 3.0.0 (see https://github.com/enthought/enable/pull/215)

corranwebster:
@itziakos is this good to merge now that tests are passing.

itziakos:
It has been a while since the tests run. I will force another build and if all pass we should be good to go

itziakos:
good to go :+1:

PR #287, opened 2016-01-28
Fix selection alpha for line scatterplot

corranwebster:
Copy-paste error from scatterplot code, plus missing trait.

### DIFF
--- a/chaco/line_scatterplot_1d.py
+++ b/chaco/line_scatterplot_1d.py
@@ -50,6 +50,9 @@ class LineScatterPlot1D(Base1DPlot):
     #: The line dash style of the selected line.
     selected_line_style = LineStyle("solid")
 
+    #: The fade amount for unselected regions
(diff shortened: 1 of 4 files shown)


jonathanrocher:
LGTM. Does this mean that we are missing a unit test to exercise the `_render` method?

jwiggins:
This same missing trait is referenced in `ScatterPlot1D` too.

corranwebster:
@jonathanrocher We were exercising `_render`, but not the branch that drew selections.  Tests are now added to do that (although they are really just smoke-tests).

@jwiggins Thanks. Fixed.

corranwebster:
I am going to merge this today unless there are objections, as the fix is needed in a consulting project.

jwiggins:
LGTM

jonathanrocher:
@corranwebster This was missing an entry into the changelog. Mind adding that in a different PR?

PR #293, opened 2016-03-30
Implemented single_axis option as described in documentation.

dianeoyen:
The single_axis option was not implemented previously, although the option exists and is described by documentation.


cfarrow:
Thanks for the patch. Does this do the right thing when the plot orientation is vertical?

PR #298, opened 2016-05-19
Fix timescale tests and calendar timescales

corranwebster:
This PR fixes a couple of long-standing bugs in the ticking code for time axes.

* The first should fix the selection of tick times for day of month and month of year, which had timezone issues.
* The second fixes the _tests_ for the time_scale module, which worked if you ran them with a computer set to UTC, but otherwise probably not.
* Adds versions of certain interesting tests which simulate running the code in non-UTC timezones, in particular timezones with non-whole-hour offsets.

For the PR reviewer: Chaco's time scale handling works internally in seconds since the Unix Epoch, but displays and selects ticks according to those times in the current timezone.

There are likely many other issues with the time scale code, but this at least allows us to run the tests cleanly and reproducably.

### DIFF
--- a/chaco/scales/tests/test_time_scale.py
+++ b/chaco/scales/tests/test_time_scale.py
@@ -1,61 +1,241 @@
 
 
 import time
+import datetime
+import os
+import contextlib
 
 import numpy as np
 
-from chaco.scales.time_scale import tfrac, trange, \
-        TimeScale, CalendarScaleSystem
+from chaco.scales.time_scale import (
+    tfrac, trange, TimeScale, CalendarScaleSystem)
 from chaco.scales.api import TimeFormatter
+
+# Note on testing:
+# Chaco assumes times are in UTC seconds since Posix Epoch but does ticking
+# in whatever the local time is on the host maschine.  This is problematic for
+# testing, because correct responses depend on current state of the test
+# machine, and will vary by location and time of year.  For hour/day/year
+# testing, where TZ matter, we select 3 timezones to test in:
+#
+# * UTC, because it's easy to reason about what the answer should be
+# * Australia/North (ie. Alice Springs, Darwin) which is UTC+9:30 and with no
+#   daylight savings time shifts, which should hopefully reveal bugs with
+#   non-integer hour offsets.
+# * Pacific/Honolulu which is UTC-10:00 and with no daylight savings time
+#   shifts, which should hopefully reveal bugs with previous-day or -year
+#   baselines.
+#
+# Note that this isn't quite bullet-proof, as things will break if Hawaii or
+# the Northern Territory ever change their timezones (eg. by introducing
+# daylight savings).  We also rely on os.environ['TZ'] working to run the
+# tests, but at the time of writing it remains to be seen if this works for
+# Windows.
+
+#----------------------------------------------------------------
+# Utilities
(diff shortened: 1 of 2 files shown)


corranwebster:
And we seem to be having problems with Enable in the test suite.

corranwebster:
See https://github.com/enthought/enable/issues/229

corranwebster:
Issues with Enable are resolved.  @cfarrow could you review or find someone to review this.

chaco/scales/time_scale.py
```diff
@@ -285,48 +285,41 @@ def ticks(self, start, end, desired_ticks=None):
...
-            end = datetime.fromtimestamp(end)
+            end_dt = datetime.fromtimestamp(end)
         except ValueError:
-            end = datetime(MAXYEAR, 1, 1, 0, 0, 0)
+            end_dt = datetime(MAXYEAR, 1, 1, 0, 0, 0)
 
+        # get range of years of interest
+        # add 2 because of python ranges + guard against timezone shifts
+        # eg. 20000101 -> 19991231 because of local timezone, so end is 1999+2
+        years = range(start_dt.year, end_dt.year+2)
```
cfarrow: Any `datetime` constructed with `MAXYEAR + 2` will raise a `ValueError`. We should guard against that here, `min(end_dt.year + 2, MAXYEAR)`, or on line 292.

corranwebster: Good catch.

chaco/scales/tests/test_time_scale.py
```diff
@@ -1,61 +1,240 @@
...
+#
+# * UTC, because it's easy to reason about what the answer should be
+# * Australia/North (ie. Alice Springs, Darwin) which is UTC+9:30 and with no
+#   daylight savings time shifts, which should hopefully reveal bugs with
+#   non-integer hour offsets.
+# * Pacific/Honolulu which is UTC-10:00 and with no daylight savings time
+#   shifts, which should hopefully reveal bugs with previous-day or -year
+#   baselines.
+#
+# Note that this isn't quite bullet-proof, as things will break if Hawaii or
```
cfarrow: From what I've read, using the `TZ` environment variable is unreliable in any context. I tested this on Windows, and it doesn't work.

AFAICT, this only seems to affect the `fromtimestamp(0.0)` below. If I'm right, then we can hardcode the `EPOCH` for each timezone that we care about and not worry about the environment variable.

corranwebster: Hmmm... that will probably work.  I was concerned that there may be other calls to the underlying datetime code that rely on system time.

On the other hand, when it is working on OS X and Travis CI as intended (even if it's not 100% portable or reliable) that makes me somewhat happier about the robustness of the code, since it protects against future errors introduced if people use something which relies on `datetime` incorrectly.

So maybe the right fix is to skip these tests on windows?

cfarrow: I just ran the tests on Windows with hardcoded Alice Springs TZ. The related tests passed, so I feel like this is better than skipping the tests on Windows.

corranwebster:
I think all issues are addressed.  Is this good to merge, @cfarrow?

cfarrow:
Thanks. :shipit:

PR #299, opened 2016-05-19
[WIP] Fix Enable in travis Build

corranwebster:
Something has gone wrong with installing Enable from github via pip.


corranwebster:
Problem looks like it was in Enable.

PR #301, opened 2016-06-06
Clean-up of Python 3 PR

corranwebster:
This is a cleanup of #261.


corranwebster:
Tests are passing, but not ready to merge yet: 2to3 was run on the examples and so there's some manual back-porting that's needed (mainly `from __future__ import print_function`, I think).

chaco/color_mapper.py
```diff
@@ -405,7 +404,7 @@ def _map(self, X):
 
         """
 
-        if type(X) in [IntType, FloatType]:
+        if type(X) in [int, float]:
```
mdickinson: Orthogonal to the PR (so please don't change it here), but it bugs me that this isn't `if instance(X, (int, float, long))`. Is it worth opening an issue to remind us to fix these explicit type checks? Or should we just wait for it to cause problems in something real?

.travis.yml
```diff
@@ -1,18 +1,21 @@
 language: python
 sudo: false
 python:
-  - 2.6
   - 2.7_with_system_site_packages
+  - 3.4
```
mdickinson: Remind me: are we explicitly unsupporting Python 3.3, or is it just inconvenient to include it in the Travis versions?

chaco/color_mapper.py
mdickinson: Though the right check is probably just an `np.isscalar(X)`.

.travis.yml
corranwebster: Travis infrastructure.  There's nothing that should stop it working on Python 3.3 I think, although if it did pose any difficulty then I'd be happy to drop support for 3.3 at this point.

examples/tutorials/scipy2008/custom_tool.py
```diff
@@ -12,14 +12,14 @@ class CustomTool(BaseTool):
     event_state = Enum("normal", "mousedown")
 
     def normal_mouse_move(self, event):
-        print "Screen:", event.x, event.y
+        print("Screen:", event.x, event.y)
```
mdickinson: Missing a `from __future__ import print_function` in this module?

examples/tutorials/scipy2008/custom_tool_click.py
```diff
@@ -10,13 +10,13 @@
 class CustomTool(BaseTool):
 
     def normal_mouse_move(self, event):
-        print "Screen point:", event.x, event.y
+        print("Screen point:", event.x, event.y)
```
mdickinson: Missing `print_function` import.

corranwebster: Yep - thank you, I thought I'd caught all of those.

mdickinson:
Changes LGTM. One question about 3.3, and two places where there's a `print_function` import missing (unless you want to print tuples in Python 2).

chaco/color_mapper.py
corranwebster: And looking at the comments, we may be able to just remove the whole method if it is in fact no longer used.

.travis.yml
mdickinson: Agreed. So long as it's no additional effort to support 3.3, I think we should do so. As soon as it becomes an effort, let's drop support.

corranwebster:
Fixes made.  Will merge assuming tests pass cleanly unless there are objections

mdickinson:
LGTM!

PR #314, opened 2016-10-06
Remove special-casing of powers of 2

corranwebster:
Code was bad and buggy and not well motiviated.

@mdickinson care to review?


mdickinson:
Maybe we should keep this code somewhere? I'm sure it has educational value, of a sort.

But not here. LGTM

PR #316, opened 2016-10-14
Appveyor setup

rahulporuri:


appveyor.yml
```diff
@@ -0,0 +1,30 @@
...
+
+init:
+    - ps: $Env:sdkbin = "C:\Program Files\Microsoft SDKs\Windows" + $Env:sdkver + "\Bin"
+    - ps: $ENV:sdkverpath = "C:\Program Files\Microsoft SDKs\Windows" + $Env:sdkver + "\Setup\WindowsSdkVer.exe"
+    - ps: $ENV:path = $Env:python + ";" + $env:python + "\scripts;" + $Env:sdkbin + ";" + $Env:path
+
+install:
+    - ps: if ((Test-Path "C:\temp") -eq 0)  {mkdir C:\temp}
+    - ps: python -m pip install --upgrade --no-binary wheel pip
+    - ps: pip install --upgrade wheel
```
itziakos: I think that we do not need to do this since you already upgrade wheel in the previous line

appveyor-install-deps.cmd
```diff
@@ -0,0 +1,6 @@
+"%sdkverpath%" -q -version:"%sdkver%"
+call setenv /x64
+
+pip install --cache-dir c:/temp numpy
+pip install --cache-dir c:/temp distribute
+pip install --cache-dir c:/temp enable
```
jvkersch: I think you should install enable from git here, similar to what is done for Travis.

PR #317, opened 2016-10-18
Largest triangle three buckets downsampling

corranwebster:
Implements the "largest triangle three buckets" downsampling algorithm for lineplots.  This currently assumes:

* index and values are floats (ie. C doubles)
* index is monotone increasing

Only gets invoked if `use_downsampling` trait is `True`.

For the algorithm, see: Sveinn Steinarsson, "Down Sampling Time Series for Visual Representation," Master's Thesis, University of Iceland, 2013. http://skemman.is/handle/1946/15343


corranwebster:
This is ready for a review.  @andrewcollette, @mdickinson and/or @jvkersch ?

chaco/downsample/_lttb.pyx
```diff
@@ -0,0 +1,93 @@
...
+            count = 0
+            next_bucket_end = <Py_ssize_t>((i+2)*bucket_size) + 1
+            if next_bucket_end > data_length:
+                next_bucket_end = data_length
+
+            for j in range(current_bucket_end, next_bucket_end):
+                count += 1
+                avg_x += (points[j, 0] - avg_x)/count
+                avg_y += (points[j, 1] - avg_y)/count
+                j += 1
```
jvkersch: This line is probably superflous...

chaco/lineplot.py
```diff
@@ -325,34 +328,19 @@ def _gather_points(self):
 
     def _downsample(self):
         if not self._screen_cache_valid:
-            self._cached_screen_pts = [self.map_screen(p) for p in self._cached_data_pts]
-            self._screen_cache_valid = True
+            m = self.index_mapper
+            delta_screen = (m.high_pos - m.low_pos)
```
jvkersch: Nit: maybe cast to int here, instead of in the invocation of `lltb`.

corranwebster:
Hrm... doesn't compile on windows due to lack of `INFINITY` identifier.

corranwebster:
Reports from actual use are that this works and is useful.  Any problems beyond this can be fixed by bugfix or enhancement PRs. Merging.

PR #320, opened 2016-10-27
MAINT: Remove all mentions of unittest2.

jvkersch:
Drive-by PR to remove all uses of `unittest2`, given that the lowest supported Python version is 2.7.


jvkersch:
@itziakos Do you have any idea what might be going on here? This is the PR that I mentioned was timing out during `python setup.py develop`. The Travis build log showed that this consistently happened during the enable build, more precisely while installing `six`. Manually doing `pip install six` before the enable build seems to fix the issue...

corranwebster:
@jvkersch if six isn't already installed, then Enable's setup.py will probably try to install via easyinstall/setuptools (I think) rather than pip.  So possibly some issue with setuptools and/or installing six via setuptools?

Enable has a rather nasty setup.py system, so it wouldn't shock me if there is something bad there.

jvkersch:
That's possible. This is the line where it goes wrong in the [build log](https://travis-ci.org/enthought/chaco/jobs/171148984#L1151):
```
Running six-1.10.0/setup.py -q bdist_egg --dist-dir /tmp/easy_install-ctwu5nzk/six-1.10.0/egg-dist-tmp-lic6svep
```
Shall we just add six to Enable's Travis requirements?

corranwebster:
Yes.

PR #321, opened 2016-11-01
DOC: Changelog entries.

jvkersch:
This adds entries in the changelog for the PRs that were merged since 4.5.0.

PR #323, opened 2016-11-02
FIX: Spectrum demo occasionally crashes with IOError

andrewcollette:
While running on OS X, the `spectrum.py` demo will occasionally crash inside PyAudio with `IOError: [Errno Input overflowed] -9981`.  This patch simply restarts the audio stream if that happens.

### DIFF
--- a/examples/demo/advanced/spectrum.py
+++ b/examples/demo/advanced/spectrum.py
@@ -104,6 +104,9 @@ def get_audio_data():
     try:
         audio_data  = fromstring(_stream.read(NUM_SAMPLES), dtype=short)
     except IOError as e:
+        # Workaround "Input overflowed" issue on OS X, by restarting stream
+        if e.errno != "Input overflowed":
+            raise
         audio_data = zeros((NUM_SAMPLES,))
         _stream.stop_stream()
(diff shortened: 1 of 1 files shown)


corranwebster:
Probably want a comment in the `except` with a brief explanation of what you're doing.

Also, if it's always error number -9981 that you're seeing, you might want to check for that explicitly and raise otherwise to avoid masking other problems.

andrewcollette:
OK, I have added a comment and check to prevent it from responding to a generic IOError.

corranwebster:
LGTM

andrewcollette:
Awesome.  Shall I merge directly into master or are there other steps to take?

corranwebster:
Yes, it can be directly merged, but check with @jvkersch about _when_ he wants it merged: he is preparing to cut a release.  It would be good to get this in, but not essential.

If it is merged, then an appropriate entry should to be added to the changelog PR https://github.com/enthought/chaco/pull/321 before merging it.

jvkersch:
Let's merge it now (I had my eye on it on Friday but wanted to hold off until Corran was back). I'll add the entry to #321.

PR #325, opened 2016-11-16
[WIP] Non-uniformly sampled 2d plots

corranwebster:
Includes fast 2d downsampling in Cython, which may be of use to other plots as well.

Work needs to be done on integration, particularly the Plot classes insistence on resampling indices, plus disambiguation of cell-based vs. vertex based data.

Plus unit tests.


jonathanrocher:
I haven't looked at the details, but noticed that you have checked in the .c file. I am assuming that's unintentional since it is generated, so I thought I would drop a note.

corranwebster:
It's intentional: for Enable and Chaco we include the .c file so that things can be built without having to have Cython installed.

Developers have to be good and remember to update the .c file if they make changes to the Cython code.

jwiggins:
What about requiring Cython for developers tracking the master branch, but including the .c file in the sdist?

jonathanrocher:
+1 for @jwiggins 's suggestion.

corranwebster:
That's a reasonable suggestion, but probably not as part of this PR.

jonathanrocher:
Agreed.

PR #326, opened 2016-11-16
Fix contour plot grids

### LINKED ISSUE #310: Contour plot problem with arrays with nan values
There seems to be an issue when using the contour plot on an array with nan values (numpy.nan). Rather than terminating the contours at the edges of a nan value it forces the contours into closed loops, often with quite erroneous results. Below is an example shown with a data set that is undefined around all edges and has a non-rectangular shape.

```
from traits.api import HasStrictTraits, Instance
from traitsui.api import View, UItem
import numpy as np


class ImgPlot(HasStrictTraits):
    plot = Instance(Plot)

    traits_view = View(
        UItem('plot', editor=ComponentEditor(size=(900,800))),
        resizable=True
    )

    def __init__(self, data_array, xbounds, ybounds):
        super(ImgPlot, self).__init__()

        # Define plot data and plot instance
        plot_data = ArrayPlotData(z=data_array)
        plot = Plot(plot_data)

        # Make image plot
        plot.img_plot('z', name='img_plot', hide_grids=False, xbounds=xbounds, ybounds=ybounds)

        # Make contour plot
        plot.contour_plot('z', type='line', name='con_plot', hide_grids=False, 
                            xbounds=xbounds, ybounds=ybounds, levels=20)
        self.plot = plot
```

picture

Unfortunately the plotting methods can't handle numpy masked arrays either.

corranwebster:
Fixes #310 amongst other issues.

This PR fixes a number of issues with contour plots, and expands their capabilities:

* allows grid points to be non-uniformly sampled x and y values (ie. not a linspace)
* allows regions to be masked out
* treats non-finite values as being masked
* fixes assumption in `Plot` code that x and y values for index must be linearly spaced
* distinguishes between 2d data plotted at vertices and 2d data plotted in cells in `Plot` code
* significant simplifications in code paths by avoiding having to handle off-by-one errors in various places.

screen shot 2016-11-16 at 3 48 05 pm

### DIFF
--- a/chaco/contour_line_plot.py
+++ b/chaco/contour_line_plot.py
@@ -4,7 +4,7 @@
 from __future__ import with_statement
 
 # Major library imports
-from numpy import array, linspace, meshgrid, transpose
+from numpy import array, isfinite, linspace, meshgrid, transpose
 
 # Enthought library imports
 from enable.api import LineStyle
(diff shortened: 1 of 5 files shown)


jvkersch:
LGTM

corranwebster:
Cool. Merging.

PR #329, opened 2016-12-14
 A couple of fixes to keep VS2010 compatibility and avoid a deprecate…

spichardo:
…d warnings with newer versions of numpy

The first fix is on chaco\contour\cntrc.c to move the moduledef
declaration to the top the MOD_INIT function to be complain with older
compilers followingn the old C89 specification (such as VS 2010), the
credit to spot this goes to @stevenengler . The change should be
compatible with any other compiler.

The second change is on chaco\base.py to avoid a deprecate warning on
newer versions of numpy that do not accept differantion of boolean
arrays. The current fix simply recasted the array as uint8.

### DIFF
--- a/chaco/base.py
+++ b/chaco/base.py
@@ -7,7 +7,7 @@ from math import radians, sqrt
 
 # Major library imports
 from numpy import (array, argsort, concatenate, cos, diff, dot, empty, isfinite,
-                   nonzero, pi, searchsorted, seterr, sin)
+                   nonzero, pi, searchsorted, seterr, sin, uint8)
 
 # Enthought library imports
 from traits.api import CArray, Enum, Trait
@@ -220,7 +220,7 @@ def arg_true_runs(bool_array):
     """ Find runs where array is True """
     if len(bool_array) == 0:
         return []
-    runs = arg_find_runs(bool_array, 'flat')
+    runs = arg_find_runs(bool_array.astype(uint8), 'flat')
     # runs have to alternate true and false
     if bool_array[0]:
         # even runs are true
--- a/chaco/base.py
+++ b/chaco/base.py
@@ -7,7 +7,7 @@ from math import radians, sqrt
 
 # Major library imports
 from numpy import (array, argsort, concatenate, cos, diff, dot, empty, isfinite,
-                   nonzero, pi, searchsorted, seterr, sin, uint8)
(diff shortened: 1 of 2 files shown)


chaco/base.py
```diff
@@ -220,7 +220,7 @@ def arg_true_runs(bool_array):
     """ Find runs where array is True """
     if len(bool_array) == 0:
         return []
-    runs = arg_find_runs(bool_array, 'flat')
+    runs = arg_find_runs(bool_array.astype(uint8), 'flat')
```
corranwebster: Possibly better to use `view` rather than `astype` to avoid a copy?

corranwebster:
Thanks for the contribution.  One minor comment, but other than that, LGTM.

Currently the unit tests are failing (it looks like it is unrelated - Travis CI tests are using master of TraitsUI but not Pyface, which is a bad idea), so we need to get those working before we can merge.

corranwebster:
I have opened an issue for the test failures #330.

corranwebster:
@spichardo Could you please merge current master into your branch - Travis CI tests should pass if you do that.

spichardo:
yes, , using view is more adequate if we can assume that `np.bool` size is always 1 byte and it never changes among versions and platforms.  `astype` can be used as  less efficient but it will always return the same matrix shape as expected. I'll change to `view` as it is reasonable to expect bool type will rarely change in size. Also I should have used instead signed `int8` instead of `uint8` as the conversion makes True==1 and False==0, then it can be expected to have '-1' in the differentiation.  I understand that at end the code wants to detect the locations where the boolean array changes values. I just made another commit for this.

Now, and I sincerely apologize in advance since I don't have yet too much experience in merging & forking,  I re-merged enthought/chaco:master all the way to my experimental branch, hopefully now the tests will pass. Please let me know if I need to do something else. cheers

corranwebster:
Yep - that looks like everything is working.  There's a good chance we'll be able to slip this into the upcoming 4.6.0 release.

Thanks for the contribution!

spichardo:
Glad to help even a tiny bit, I've been happily using chaco+enable+traitsui for years now in our applications. Until recently I never had to use the development versions since we stayed in the cozy world of Python2.7. But we want know to certify our applications and libraries can do the jump to Python3.X so then I had to start using the development versions.

PR #328, opened 2016-12-02
Release 4.6.0

corranwebster:
Release branch for 4.6.0

PR #332, opened 2016-12-15
Correct dtype for contour mask.

jvkersch:
Fix for the issue described in https://github.com/enthought/chaco/pull/328#issuecomment-264516430

@corranwebster Could you give this a review? I added some tests for the Contour class, too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ addons:
     - libjpeg8-dev
     - zlib1g-dev
     - libpng-dev
-    - libfreetype36-dev
+    - libfreetype6-dev
     - python-cairo
     - cmake
     - swig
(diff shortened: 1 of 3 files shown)


corranwebster:
Something weird is going on with Travis CI, I think.

jvkersch:
@corranwebster I don't know what happened there, but updating libfreetype36-dev to a version of the package that I could find on packages.ubuntu.com, namely libfreetype6-dev, seems to have fixed the problem.

corranwebster:
I am nervous about the change, since it is possible that we are getting silent failures on text drawing (eg. warnings about not finding fonts) because of library differences, but I am willing to live with that for the time being.

Other than that, LGTM (especially the tests)!

PR #337, opened 2016-12-23
fix variable names for building from source distribution

grlee77:
I received the following error when trying to build from the source 4.6.0 tarball obtained from
https://github.com/enthought/chaco/releases/download/4.6.0/chaco-4.6.0.tar.gz
```
  File "setup.py", line 133, in <module>
    write_version_py()
  File "setup.py", line 113, in write_version_py
    match = re.match(r'.*?\.dev(?P<dev_num>\d+)', full_v)
NameError: name 'full_v' is not defined
```

This PR fixes the variable names in this section of setup.py.
Not 100% certain if it should be `full_version` or `fullversion`, but I went with the latter as that is what seemed to be used in the other cases.

(Alternately, deleting chaco/_version.py also allows the build to proceed by avoiding this section of setup.py)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -101,14 +101,14 @@ if not is_released:
     elif os.path.exists(filename):
         # must be a source distribution, use existing version file
         try:
-            git_revision, full_version = read_version_py(chaco_version_path)
+            git_rev, fullversion = read_version_py(chaco_version_path)
         except (SyntaxError, KeyError):
             raise RuntimeError("Unable to read git_revision. Try removing "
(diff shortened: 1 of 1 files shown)


corranwebster:
Thanks for the contribution! This was entirely my fault copy pasting code to get the release out and avoid a different bug in this section.

We'll probably put out a 4.6.1 release shortly to fix the issue.

mdickinson:
@corranwebster: Did the 4.6.1 release ever make it to PyPI?

PR #342, opened 2017-03-08
BUG: Copy over HOME value to git environment.

rkern:
Some installations of git have `~` expansions in their global `gitconfig` files. `git` processes need to have the `HOME` environment variable defined.

This fix will need to be propagated to other `setup.py` files that reuse this code.


jvkersch:
LGTM

PR #350, opened 2017-04-11
Use ArrayOrNone

rkern:
Fixes all of the instances of ```FutureWarning: comparison to `None` will result in an elementwise object comparison in the future.``` that I could find.

Additionally clean up a few warnings so that we have a clean test suite.

### DIFF
--- a/chaco/tests/test_array_or_none.py
+++ b/chaco/tests/test_array_or_none.py
@@ -16,6 +16,8 @@ from chaco.scatterplot import ScatterPlot
 class TestArrayOrNone(unittest.TestCase):
     """ Test that the FutureWarning from numpy concerning comparison of arrays
     against None are not issued.
+
+    These warnings began with numpy 1.9.
     """
 
     def test_array_data_source(self):
--- a/chaco/tests/test_array_or_none.py
+++ b/chaco/tests/test_array_or_none.py
@@ -1,0 +1,60 @@
+import unittest
+import warnings
+
+import numpy as np
+
+from chaco.array_data_source import ArrayDataSource
+from chaco.axis import PlotAxis
+from chaco.data_label import DataLabel
+from chaco.data_range_1d import DataRange1D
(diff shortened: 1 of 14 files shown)


chaco/tools/range_selection.py
```diff
@@ -133,7 +133,7 @@ class RangeSelection(AbstractController):
 
     # The data space start and end coordinates of the selected region,
     # expressed as a list.
```
cfarrow: This comment is out of date.

chaco/tests/test_array_or_none.py
```diff
@@ -0,0 +1,60 @@
...
+from chaco.data_label import DataLabel
+from chaco.data_range_1d import DataRange1D
+from chaco.label_axis import LabelAxis
+from chaco.legend import Legend
+from chaco.linear_mapper import LinearMapper
+from chaco.scatterplot import ScatterPlot
+
+
+class TestArrayOrNone(unittest.TestCase):
+    """ Test that the FutureWarning from numpy concerning comparison of arrays
```
cfarrow: It would be helpful to expand upon this to include the numpy version where this is a warning and the version where it will become an exception.

cfarrow:
The test failure is on python 3.4 using. It appears to be due to a setup issue with pyside (perhaps [this](https://github.com/PySide/pyside2-setup/issues/11)). I've not looked deeper into it.

rkern:
@cfarrow The remaining failures seem to be general problems with the Travis setup.

cfarrow:
Yes. The test failures look unrelated. We can deal with them elsewhere.

cfarrow:
LGTM

rkern:
Note that the last change to get this working with Enable's `master` due to enthought/enable/pull/250 makes Chaco depend on Enable `master`. I don't know if that changes your LGTM.

cfarrow:
The next Chaco release will require an Enable release. I'm not too worried about that. Still LGTM

PR #360, opened 2017-06-27
Update raise stmt

rahulporuri:


chaco/plot.py
```diff
@@ -173,8 +173,8 @@ def __init__(self, data=None, **kwtraits):
             elif type(data) in (ndarray, tuple, list):
                 self.data = ArrayPlotData(data)
             else:
-                raise ValueError, "Don't know how to create PlotData for data" \
-                                  "of type " + str(type(data))
+                raise ValueError("Don't know how to create PlotData for data"
+                                 "of type {!s}".format(type(data)))
```
jvkersch: You probably don't need the explicit `!s` specifier here.

chaco/plot.py
```diff
@@ -173,8 +173,8 @@ def __init__(self, data=None, **kwtraits):
             elif type(data) in (ndarray, tuple, list):
                 self.data = ArrayPlotData(data)
             else:
-                raise ValueError, "Don't know how to create PlotData for data" \
-                                  "of type " + str(type(data))
+                raise ValueError("Don't know how to create PlotData for data"
+                                 "of type {}".format(type(data)))
```
mdickinson: Do we care about Python 2.6? I assume not, but we should probably make that explicit somewhere.

(If we do care about Python 2.6, this needs to be `{0}` rather than `{}`.)

mdickinson: While you're here, you could also fix the missing space between "data" and "of".

rahulporuri: we only test with 2.7 but I guess the README is where this information belongs best?

README.rst
```diff
@@ -42,6 +42,7 @@ Features
 
 Prerequisites
 -------------
+Chaco is only supported on Python 2.7.x and Python > 3.3.
```
jvkersch: Maybe `Python >= 3.4` is more clear, or something like `Python 3.4 and up`?

mdickinson:
LGTM. @jvkersch?

jvkersch:
LGTM too. The Travis failures seem unrelated.

rahulporuri:
merging.

PR #365, opened 2017-07-09
EDM-based CI

jvkersch:
Cleaned-up version of #364 

This adapts the EDM-based CI code for Enable in https://github.com/enthought/enable/pull/281 to Chaco. The biggest change is to run the backend tests in addition to the standard tests (however, judging from the [README](https://github.com/enthought/chaco/blob/master/chaco/tests_with_backend/README.txt) we could probably move the backend tests back in with the rest of the tests).

@corranwebster @itziakos @jwiggins Would you mind taking a look?


corranwebster:
Basically looks good to me.  Obvious improvements would be:

- minimizing the number of different runtimes installed (ie. install a single 2.7 runtime with wx and pyside and then run tests for each).  Doing this improved runtimes and lowered cache sizes.
- running on OS X (should be straightforward)
- setting up Appveyor (more work)

but those are all second order compared to getting this into master.

jwiggins:
I would remove all the tomfoolery concerning pillow in `edmtool.py` since it's not needed.

.travis.yml
```diff
@@ -1,61 +1,58 @@
...
-    - python: 3.5
-      env: ETS_TOOLKIT='qt4'
-cache:
-  - pip
-  - ccache
+
 addons:
   apt:
     packages:
     - python-qt4
```
itziakos: since we use the edm packages we should not need to install 
python* packages and zlibg-dev, libpng-dev, libfreetype6-dev  from apt

.travis.yml
```diff
@@ -1,61 +1,58 @@
...
+
+env:
+  global:
+    - INSTALL_EDM_VERSION=1.5.2
+      PYTHONUNBUFFERED="1"
+
+matrix:
+  include:
+  - env: RUNTIME=2.7 TOOLKIT=wx
+  - env: RUNTIME=2.7 TOOLKIT=pyqt
```
itziakos: Is there a reason why pyside is not included in the matrix jobs?

ci/edmtool.py
```diff
@@ -0,0 +1,302 @@
...
+reliability of tests be removing many of the variables around the
+developer's particular Python environment.  Test environment setup and
+package management is performed using `EDM
+http://docs.enthought.com/edm/`_
+To use this to run you tests, you will need to install EDM and click
+into your working environment.  You will also need to have git
+installed to access required source code from github repositories.
+You can then do::
+    python edmtool.py install --runtime=... --toolkit=...
+to create a test environment from the current codebase and::
```
itziakos: Adding some empty lines will include readability in these sections

.travis.yml
jvkersch: Good call, thanks. Updated.

.travis.yml
jvkersch: No particular one, other than that I looked at the current Travis file and misread what it does. I'll add PySide.

jvkersch: I've added a configuration for PySide under 2.7.

jvkersch:
@jwiggins @itziakos I think I've addressed your comments, would you care to give this another look?

jwiggins:
Looks good from this end

jvkersch:
Thanks all, merging.

@corranwebster I'll take a shot at OS X and Windows soon.

jwiggins:
Nice work! Any insights to be added to enthought/enable#281?

jvkersch:
> Any insights to be added to enthought/enable#281?

@jwiggins Just one perhaps: Ioannis pointed out that installing the python-* packages via apt is no longer necessary, so it looks like those could be removed from the Enable setup as well. Other than that, nothing to add: the Python code is copied in its entirety from your branch.

Issue #366, opened 2017-07-15
ZoomTool doesn't play well with secondary axis


jonathanrocher:
The ZoomTool doesn't work well with the BroadcasterTool and multiple axis in a Plot. Running the `multiaxis_using_Plot.py` example, and zooming in a certain range leads to a new plot with the regular axis has the correct range but not the secondary axis.

<img width="901" alt="screen shot 2017-07-15 at 4 54 03 pm" src="https://user-images.githubusercontent.com/593945/28242926-ca1a52fc-697e-11e7-8391-ca192b123927.png">

From the full plot, if I zoom say from (0, 0) to (2, 0.2), the y-axis range should roughly be [-0.05, 0.05]. Instead I get the following:

<img width="897" alt="screen shot 2017-07-15 at 4 57 54 pm" src="https://user-images.githubusercontent.com/593945/28242925-ca17c366-697e-11e7-97a5-cc9dd5bf0076.png">

jonathanrocher:
cc @rkern @cfarrow .

jtang-ent:
I had similar problem and found out that I can't share the same value mapper for different y-axis plot in order for zooming to work properly. If you create a container first and then add both `plot1` and `foreign_plot` to the container the zooming should work as expected.

jonathanrocher:
Looking at this again, it seems like the `ZoomTool` bad box selection behavior is also exhibited when used on a trivial `OverlayPlotContainer`, independently of the `BroadcasterTool`:
```
import numpy as np
from traits.api import HasTraits, Instance
from traitsui.api import Item, View
from chaco.api import add_default_axes, add_default_grids, create_line_plot, \
    OverlayPlotContainer
from chaco.tools.api import PanTool, ZoomTool
from enable.api import ComponentEditor

class PlotView(HasTraits):
    plot = Instance(OverlayPlotContainer)

    view = View(
        Item("plot", show_label=False, editor=ComponentEditor(), style="custom")
    )

    def _plot_default(self):
        x = np.linspace(0, 2*np.pi)
        y = np.sin(x)
        container = OverlayPlotContainer(padding=50)

        plot = create_line_plot((x, y))
        add_default_axes(plot)
        add_default_grids(plot)
        container.add(plot)

        container.overlays.append(ZoomTool(component=plot))
        container.tools.append(PanTool(component=plot))
        return container

if __name__ == "__main__":
    PlotView().configure_traits()
```

PR #369, opened 2017-07-20
Python 3 compatibility fixes

jdeschenes:
The unit test pass in python 3.6. **This has not yet been completely tested in python**. I will do another pass once I have gone over enable.

The examples have not been run as well. So I may make further modifications.

There is an issue in python 3 in the image plot. This is caused by a bug in enable. I skipped those tests.


chaco/legend.py
```diff
@@ -355,7 +358,7 @@ def get_preferred_size(self):
         if len(self.plots) == 0:
             return [0, 0]
 
-        plot_names, visible_plots = map(list, zip(*sorted(self.plots.items())))
+        plot_names, visible_plots = list(sm.map(list, sm.zip(*sorted(six.iteritems(self.plots)))))
```
jvkersch: This could probably do with some unpacking. Alternatively, can we just use the standard Python constructs here and write the RHS as `list(map(list, zip(*sorted(self.plots))))`?

chaco/plotscrollbar.py
```diff
@@ -1,4 +1,6 @@
```
jvkersch: Can you put a `from __future__ import print_function` here?

chaco/scales/formatters.py
```diff
@@ -350,15 +354,19 @@ def estimate_width(self, start, end, numlabels=None, char_width=None,
             elif char_width:
                 avg_size = len("%g%g" % (start, end)) / 2.0
                 initial_estimate = round(fill_ratio * char_width / avg_size)
+            else:
```
jvkersch: This seems orthogonal to the cleanup, can you describe what the issue is here? If the issue is that `numlabels` and `char_width` should not be both `None`, perhaps a `ValueError("num_labels and char_width should not both be None.")` would be more appropriate.

chaco/scales/scales_test_case.py
```diff
@@ -3,8 +3,8 @@
```
jvkersch: Also needs a `from __future__ import print_function`.

chaco/scales/tests/test_formatters.py
```diff
@@ -42,7 +42,7 @@ def test_strftimeEx_04():
     # The format "%(ms_)" uses floor().
```
jvkersch: Also needs a `from __future__ import print_function` at the top of the module.

chaco/scales/time_scale_test_case.py
```diff
@@ -2,11 +2,11 @@
 from itertools import starmap
```
jvkersch: Also needs a `from __future__ import print_function` at the top of the module.

chaco/tests/test_image_plot.py
```diff
@@ -108,48 +113,56 @@ def plot_comparison(input_image, expected_image, **plot_kwargs):
     plt.show()
 
 
+@unittest.skipIf(six.PY3, "Bug in the image plotter in python 3")
```
jvkersch: That's unexpected, is that new?

chaco/tools/line_inspector.py
```diff
@@ -78,7 +78,8 @@ def draw(self, gc, view_bounds=None):
         if self.is_listener:
             tmp = self._get_screen_pts()
         elif self.is_interactive:
-            tmp = self._last_position
```
jvkersch: It looks like this (and some of the other changes in this module) changes the functionality. Can you explain why this is necessary? If this fixes a bug, it would probably be good to split it off from the current PR and submit as a separate PR.

chaco/tools/pan_tool.py
```diff
@@ -247,6 +247,8 @@ def panning_mouse_leave(self, event):
         return self._end_pan(event)
 
     def _start_pan(self, event, capture_mouse=True):
+        if event.control_down:
```
jvkersch: Also a functionality change.

docs/source/sphinxext/comment_eater.py
```diff
@@ -1,10 +1,11 @@
-from cStringIO import StringIO
```
jvkersch: I think some of the bundled sphinx extensions in this folder might not be Python3 compatible... (Just a comment in passing, no need to take any action)

examples/demo/basic/image_lasso.py
```diff
@@ -27,15 +27,15 @@ def lasso_updated(lasso_tool, name, old, new_selections):
...
+        #print("Selection region", i)
 
         # We first map to screen because the selection is stored as coordinates
         # in data space
         screen_pts = lasso_tool.plot.map_screen(selection)
 
         # Now map each point into the grid index
-        for x, y in screen_pts:
-            print("\t", lasso_tool.plot.map_index((x, y)))
+        # for x, y in screen_pts:
```
jvkersch: Can you remove the commented-out code here?

examples/demo/image_plot_origin_and_orientation.py
```diff
@@ -13,7 +13,7 @@
 """
 
 # Major library imports
-from scipy.misc import lena
+from scipy.misc import face
```
jvkersch: +100

examples/demo/multiaxis_using_Plot.py
```diff
@@ -24,6 +24,60 @@
 from chaco.tools.api import BroadcasterTool, PanTool, ZoomTool
 from chaco.api import create_line_plot, add_default_axes
 
+from traits.api import List
+from chaco.tools.api import SelectedZoomState
+import numpy
+
+class CaeZoomTool(ZoomTool):
```
jvkersch: Cae?

jdeschenes:
Sorry about that, looks like I bundled more changes(I am using this branch) than strictly necessary. I will do a pass and remove functional changes.

chaco/tools/line_inspector.py
jdeschenes: Hi, I pulled in my branch #205 as I found it helped quite a bit and it looked like it will not be implemented anytime soon. I reverted changes in latest commit.

chaco/tools/pan_tool.py
jdeschenes: Sorry about that... This has been reverted.

examples/demo/multiaxis_using_Plot.py
jdeschenes: Sorry about that, this should not have been part of the commit.

jvkersch: No worries, looked like a pretty nifty tool ;-)

jdeschenes:
I think I need to comment on some of the functional changes that I put there by mistake.

For multiaxis_using_Plot.py:

1. Using the zoom tool with ctrl is doing a pan at the same time(pan_tool.py.) Strangely, this doesn't happen in two_plots.py. I think a code change may be necessary here. My approach was to subclass PanTool and ignore the pan move if my modifiers were pressed.
2. The Box on the ZoomTool does not appear
3. The ZoomTool is not doing the correct zoom(See issue #366) The CaeZoomTool has this fixed, but only partially... Using the mouse wheel to zoom off will not do what is right.

For financial_plot.py

1. The font does not render properly on my machine. There is a warning on font_manager.py in kiva. That's a complete separate issue.

These changes definitively don't belong in this pull request.

chaco/tests/test_image_plot.py
jdeschenes: That's actually a pretty old [issue](https://github.com/enthought/enable/issues/283)

chaco/legend.py
jdeschenes: Your approach will not work. You need to have the items() in there

```python
x = {'b': 1, 'a': 3}

assert sorted(x) == ['a', 'b']
assert sorted(x.items()) == [('a', 3), ('b', 1)]
```

For standard python contructs. See my comment in enthought/traits#374

It's more or less the same issue. I would prefer to keep the iterator version whenever possible.

Let me know your decision.

jvkersch: Ah, I misunderstood. Thanks for the clarification, and let's keep it like you had it originally.

chaco/tests/test_image_plot.py
jvkersch: I see, can you add a reference to the issue in the skip message?

jvkersch:
@jdeschenes Thanks for addressing the comments. Would you mind merging master in and resolving any merge conflicts? It would be good to start getting Travis test results on this.

jdeschenes:
The merge has just been done.

chaco/speedups.py
```diff
@@ -1,12 +1,14 @@
...
+from ._speedups_fallback import *
 
 
 # cython implementation of speedups. Import these if we can.
 try:
-    from _cython_speedups import *
+    from ._cython_speedups import *
 except ImportError:
+    # TO TEST
+    raise
```
jvkersch: Do we need this explicit `raise`?

docs/source/conf.py
```diff
@@ -52,7 +52,9 @@ def get_build_docset():
 # The default replacements for |version| and |release|, also used in various
 # other places throughout the built documents.
 d = {}
-execfile(os.path.join('..', '..', 'chaco', '_version.py'), d)
+exec(compile(open(os.path.join('..', '..', 'chaco', '_version.py')).read(),
```
jvkersch: Can you isolate the `os.path.join('..', '..', 'chaco', '_version.py')` into a local variable for re-use?

jvkersch:
@jdeschenes I left two tiny comments, but otherwise this looks very good to me. If you fix up those two items (and have nothing more to add), then I think this is good to go.

jdeschenes:
@jvkersch I fixed the issues you mentionned. I think this is ready to go!

jvkersch:
@jdeschenes I'm following along with the Enable review, and I think it would be good here too to use `.items()`, `.values()`, etc. instead of `six.iteritems(...)`, `six.itervalues(...)`, and so on. Would you be able to make that change? If not, I will pick this up separately (and apologies for bringing this up at the last minute).

jdeschenes:
@jvkersch I think I have corrected everything. Let me know if there is anything else!

jvkersch:
@jdeschenes I think this looks good to go! 

Thanks a lot for doing this (and for tackling the other repos, too). This will improve developer experience by a lot!

Issue #376, opened 2017-09-26
EPOCH definition fails on win10 py3.6


ThomasLecocq:
**Problem Description**
Hi,
I'm (almost done) adapting our seismogram picking software built on traitsui+chaco to py3.

This line:
https://github.com/enthought/chaco/blob/7c61319a20669364f67547d7a65b186e4cc9232f/chaco/scales/safetime.py#L15

fails on win10 x64 py3.6 

The minimum value acceptable is 86400

**Reproduction Steps:**

```python
In [1]: import datetime

In [2]: datetime.datetime.fromtimestamp(0)
---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
<ipython-input-2-ca3d263171b9> in <module>()
----> 1 datetime.datetime.fromtimestamp(0)

OSError: [Errno 22] Invalid argument

In [3]: datetime.datetime.fromtimestamp(86400)
Out[3]: datetime.datetime(1970, 1, 2, 1, 0)
```

**Expected behavior:**
I'd suspect that the desired behaviour is
```python
epoch = datetime.datetime(1970,1,1)
```

but then, this might fail if another "0.0" is defined ?

**OS, Python version:** 
Win10 x64 Py3.6 (Python 3.6.2 | Anaconda custom (64-bit))

jvkersch:
A description of the underlying issue is given here: https://bugs.python.org/issue29097

Not sure what the appropriate fix is, I'm not overly familiar with that part of the code. Will look at it some more soon, unless someone beats me to it.

corranwebster:
I went over the timescale code in a fairly detailed manner last year because we had tests which passed if you were in one timezone and failed if you were in another (or if daylight savings was in force...)

But no, `datetime.datetime(1970, 1, 1)` won't work: in my current TZ (BST) they give different results:
```
>>> datetime.datetime(1970, 1, 1)
datetime.datetime(1970, 1, 1, 0, 0)
>>> datetime.datetime.fromtimestamp(0)
datetime.datetime(1970, 1, 1, 1, 0)
```

I'm guessing that that the appropriate work-around is to do something gross like `datetime.datetime.fromtimestamp(86400) - 86400`

corranwebster:
Possibly need to catch `OSError` in `safe_fromtimestamp` as well.

ThomasLecocq:
Maybe check what we do in Obspy:
[...]

PR #379, opened 2017-10-05
Rel/chaco 4 7 0

jvkersch:


corranwebster:
I'm getting failures using the new edmtools CI on OS X because it isn't installing swig before trying to build the git checkout of enable/kiva.

jvkersch:
@corranwebster I haven't tested the EDM setup on Mac OS at all, the earlier goal was just to replicate the existing Ubuntu workflow. Can you open an issue for this? Also, I'd rather not hold up the release for this -- I'm thinking of doing a quick patch release in a few weeks from now once we have a fix for #376 in place.

corranwebster:
For the record, other than the problem with running the CI scripts, the PR looks fine.  Feel free to merge and CI can be fixed later.

PR #383, opened 2017-10-31
MAINT: Upcast to int_ instead of int64 to avoid bincount issue.

jvkersch:
This updates the computation of the root-mean square function (used in tests to compare generated images with the expectation) to use int_ instead of blanket 64-bit. This came up because the result of the upcast is fed into `np.bincount`, which on 32-bit Windows will raise a `TypeError` for 64-bit inputs.


jvkersch:
Thanks @itziakos

mdickinson:
> which on 32-bit Windows will raise a TypeError for 64-bit inputs

Ouch. Seriously?! Is there an open NumPy issue for that?

jvkersch:
@mdickinson This looks like it's related: https://github.com/numpy/numpy/pull/4366

PR #387, opened 2017-11-08
FIX: Avoid initializing EPOCH from t = 0.

jvkersch:
@corranwebster This is your fix from https://github.com/enthought/chaco/issues/376. Can you give this a quick look?


jvkersch:
In the absence of CI on Windows, I've tested manually that this fixes the issue on Windows 10.

PR #395, opened 2017-12-29
Feature: scatter inspector event

jonathanrocher:
Goal: I am building connected scatter plots and wanted to propagate hover and selection events from one plot to the other. Selection events can be propagated by listener to the tool's component's index's metadata_changed, but that's very labor intensive. And there was no way to propagate hover events.

This PR fixes both these issues, by proposing to emit custom `ScatterInspectorEvent`, which holds information about the type of event, and the scatter point index being involved. The logic is:
  * a `select` type event gets emitted on selection and a different `deselect` event gets emitted when deselecting a point.
  * a single `hover` event gets emitted when hovering a point (no event if the mouse moves while on the same point)
  * a `hover` event gets emitted when mouse leaves a point, with index `None`.

This PR also adds tests for the inspector that were missing, and a tiny bit of styling fixes that I couldn't refrain from making.

Backward compatibility: This PR should be 100% backward compatible, since I am leaving the (now redundant) `getattr(plot, name).metadata_changed = True` events. I thought of replacing `True` by a more interesting `ScatterInspectorEvent`, but that would require modifying the `AbstractDataSource` code, which feels too intrusive for my goal.

Feedback is welcome, especially on the strategy for when to emit hover events, and whether it would be more logical to emit some "`unhover`" events when leaving a point rather than a `hover` event with `None` index.

### DIFF
--- a/chaco/tools/scatter_inspector.py
+++ b/chaco/tools/scatter_inspector.py
@@ -2,7 +2,7 @@
 """
 
 # Enthought library imports
-from traits.api import Any, Bool, Event, HasStrictTraits, Str
+from traits.api import Any, Bool, Enum, Event, HasStrictTraits, Str
 
 # Local, relative imports
 from .select_tool import SelectTool
@@ -16,9 +16,9 @@ DESELECT_EVENT = "deselect"
 
 class ScatterInspectorEvent(HasStrictTraits):
     #: Is it a hover event or a selection event?
-    event_type = Str
+    event_type = Enum([HOVER_EVENT, SELECT_EVENT, DESELECT_EVENT])
 
-    #: What index was inspected?
+    #: What index was involved?
     event_index = Any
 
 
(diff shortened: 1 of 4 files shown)


chaco/tools/scatter_inspector.py
```diff
@@ -2,11 +2,25 @@
...
+HOVER_EVENT = "hover"
+
+SELECT_EVENT = "select"
+
+DESELECT_EVENT = "deselect"
+
+
+class ScatterInspectorEvent(HasStrictTraits):
+    #: Is it a hover event or a selection event?
+    event_type = Str
```
jvkersch: Maybe make this into an Enum for the types defined above?

jonathanrocher:
@jvkersch Thanks for the fast review. I addressed your comments. 1 minor things I want to make sure you like: I added a comment above the metadata_changed event being issued saying that it was deprecated now, and people should use the new event. If you don't agree, I can remove the comment.

jvkersch:
@jonathanrocher I think this is good to go. I agree with having the warning in the comment that `metadata_changed` can now be considered as deprecated. In a separate PR, we should switch over the consumers of `metadata_changed` to the new api, and then start deprecating the event in earnest (over several releases). I'll make a ticket for this.

PR #402, opened 2018-03-30
FIX: Respect visibility in the LegendHighlighter

jwiggins:
Before, the `LegendHighlighter` tool would respond to mouse clicks even if the legend was not visible.

### DIFF
--- a/chaco/tools/legend_highlighter.py
+++ b/chaco/tools/legend_highlighter.py
@@ -1,6 +1,5 @@
 import operator
 
-import six
 import six.moves as sm
 
 # ETS imports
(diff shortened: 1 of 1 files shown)


chaco/tools/legend_highlighter.py
```diff
@@ -1,6 +1,5 @@
 import operator
 
-import six
```
jwiggins: Unrelated cleanup. This import was not used at all in the file

corranwebster:
LGTM, although is this the only place where there is this sort of problem?  I'm not sure we should do something as drastic as not dispatching events to non-visible components in Enable, but it might be worth a quick audit of other tools.

jwiggins:
Thanks for the quick review. Shall I create an issue for the audit of the other tools?

corranwebster:
Yes, that would be a good idea.

PR #403, opened 2018-03-30
FIX: Handle multiple plots in LegendHighlighter

jwiggins:
A Legend can have multiple plot renderers per name.

### DIFF
--- a/chaco/tools/legend_highlighter.py
+++ b/chaco/tools/legend_highlighter.py
@@ -1,10 +1,18 @@
-import operator
-
-import six.moves as sm
+from itertools import chain
 
 # ETS imports
 from chaco.tools.api import LegendTool
 from traits.api import List, Float
+
+concat = chain.from_iterable
+
+
+def _ensure_list(obj):
+    """ NOTE: The Legend stores plots in a dictionary with either single
+    renderers as values, or lists of renderers.
+    This function helps us assume we're always working with lists
+    """
+    return obj if isinstance(obj, list) else [obj]
 
 
 def get_hit_plots(legend, event):
@@ -26,7 +34,7 @@ def get_hit_plots(legend, event):
         ndx = legend._cached_labels.index(label)
         label_name = legend._cached_label_names[ndx]
         renderers = legend.plots[label_name]
-        return renderers
+        return _ensure_list(renderers)
     except (ValueError, KeyError):
         return []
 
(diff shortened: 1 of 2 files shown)


chaco/tools/legend_highlighter.py
```diff
@@ -1,13 +1,16 @@
...
 from chaco.tools.api import LegendTool
 from traits.api import List, Float
 
 
+def _ensure_list(obj):
+    """ NOTE: The Legend stores plots in a dictionary with either single
+    renderers as values, or lists of renderers.
+    This function helps us assume we're always working with lists
+    """
+    return obj if isinstance(obj, list) else [obj]
```
jwiggins: I briefly looked into the code for `Legend` to see if it made more sense to fix things there, but that did not appear to be the safe approach.

chaco/tools/legend_highlighter.py
```diff
@@ -54,35 +57,34 @@ def normal_left_down(self, event):
...
+
+        if plots:
+            plots[0].request_redraw()
 
         event.handled = True
 
     def _reset_selects(self, plots):
         """ Set all renderers to their default values. """
-        for plot in sm.reduce(operator.add, plots.values()):
+        for plot in sum((_ensure_list(p) for p in plots.values()), []):
```
jwiggins: `plots` is the dictionary of the same name on the `Legend` instance. We need to anticipate `values()` being single renderers or lists of renderers.

jwiggins:
@corranwebster The example code only covers the single plot case. I have a very similar change in some product code if you'd like to see something which exercises the multi-plot case.

chaco/tools/legend_highlighter.py
corranwebster: A better construct for concatenating list-likes is `itertools.chain.from_iterable` to the point where I sometimes alias that as `concat`.

jwiggins: Yup. Done.

jwiggins:
Thanks for the feedback!

jonathanrocher:
Want to make sure I understand the fix. Before merging, selecting a name would show and hide only 1 of the renderers, am I guessing correctly? Mind adding details in the description of the issue?

jwiggins:
@jonathanrocher That's correct.

If you have a plot with a legend where, for example, you have a single logical curve/plot which is broken into multiple segments, and thus multiple renderers, the `Legend` class can handle this. However, until now the `LegendHighlighter` ignored this feature and just highlighted the first renderer returned from the `plots` dictionary. Additionally, if you _didn't_ treat `Legend.plots` as a dictionary of lists of renderers (like much of the demo code), then an exception would be raised by most methods in `LegendHighlighter`. Both of those issues were corrected here.

PR #407, opened 2018-05-07
2D Segment and Text Scatterplot types

corranwebster:
I got a bit bored on a long weekend.

screen shot 2018-05-07 at 9 15 27 pm

This PR adds a general segment plot, which is a collection of (disconnected) line segments specified by start and end points.  Segments can be mapped by width and color as well as position.

screen shot 2018-05-07 at 9 17 38 pm

Also added a plot which draws text at specified x and y locations, similar to the 1d text plot that already exists.


examples/demo/basic/minard_napoleon.py
```diff
@@ -0,0 +1,243 @@
...
+        PanTool(map_plot),
+        ZoomTool(map_plot),
+    ])
+
+    temp_plot = Plot(plot_data, height=100)
+    temp_plot.index_range = map_plot.index_range
+    temp_plot.x_grid = None
+    temp_plot.x_axis = None
+    temp_plot.y_axis.orientation = 'right'
+    temp_plot.y_axis.title = u'Temp (°C)'
```
jvkersch: I was looking at this on Wikipedia (https://en.wikipedia.org/wiki/French_invasion_of_Russia#/media/File:Minard.png), and it looks like the temperature may be in degrees Reaumur.

jvkersch:
LGTM (x2) -- can't wait to use this!

PR #421, opened 2018-07-14
Fix up create_plot_snapshots script

jvkersch:
Not sure how useful this will be for restoring the now-defunct gallery, but this fixes up the `create_plot_snapshots` script that makes snapshots of some commonly used plot types.

Also fixes a minor bug in the main codebase.


rkern:
LGTM!

PR #422, opened 2018-07-15
ENH: skip DataFramePlotData tests if pandas is not installed

### LINKED ISSUE #392: Chaco depends on Pandas for testing
The DataFramePlotData container relies on Pandas being around for testing. In the interest of keeping Pandas as an optional dependency of Chaco, we should detect at import time whether Pandas is available, and skip the DataFramePlotData tests if not.

ku-ya:
This PR adds pandas DataFramePlotData skip if pandas package does not exist.

Closes #392


ku-ya:
I will update this PR with @corranwebster 's suggestion for having skip decorator for each test case to skip.

PR #427, opened 2018-10-29
Fix float indices in Spectrum demo

achabotl:
Fixes issue where NumPy doesn't accept float values to define shapes, to index, or to slice by replacing the true divisions with floor divisions (`//`) when dividing the number of samples by 2.

```python
  File "spectrum.py", line 186, in <module>
    popup = Demo()
  File "spectrum.py", line 172, in __init__
    self.plot = _create_plot_component(self.controller)
  File "spectrum.py", line 41, in _create_plot_component
    empty_amplitude = zeros(NUM_SAMPLES/2)
TypeError: 'float' object cannot be interpreted as an integer
```


cfarrow:
I cannot test this on my system, but the changes LGTM.

PR #430, opened 2019-01-10
Add option for aligning label in index direction in TextPlot1D

jenniportman:
This is a nice little feature that I implemented for another project and wanted to port here as well. This PR adds options to align the label in the index direction (either left/center/right or top/center/bottom depending on orientation). There were already options for this sort of thing in the non-index direction but there was no way to do it in the index direction.

Screenshot with toy plot showing the new possible positioning: left/center/right or top/center/bottom (previously the label was always placed where the blue letters are)
screenshot 2019-01-10 at 11 59 38

### DIFF
--- a/chaco/text_plot_1d.py
+++ b/chaco/text_plot_1d.py
@@ -46,6 +46,10 @@ class TextPlot1D(Base1DPlot):
     #: alignment of text relative to non-index direction
     alignment = Enum("center", "left", "right", "top", "bottom")
 
+    #: alignment of text relative to index direction and centered around the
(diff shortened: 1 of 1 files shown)


corranwebster:
This looks like a useful addition. This plot has a test suite, so tests would be useful, even if they are just smoke tests that plot with each of the settings and check that the plot isn't all white.

jenniportman:
Thanks for merging this. I am sorry to admit I completely forgot about this PR and the tests you asked me to write 😞

PR #436, opened 2019-04-23
Use standard Sphinx extensions for trait documenting

### LINKED ISSUE #324: Docs don't build under Python 3
There are some Python 2-isms in `conf.py` which are easily fixed, but some of the Sphinx plugins seem like more work (e.g. `CommentEater` uses the `compiler` module that was removed under Python 3).

jvkersch:
closes #324 

This is an attempt to get docs to build under Python 3, by using standard Sphinx extensions instead of the custom ones that were bundled with Chaco (and which are Python 2 only).

This needs some scrutiny to make sure that the generated docs match the old version of the docs.


corranwebster:
LGTM

PR #446, opened 2019-06-05
Update unittest imports

rahulporuri:
Import `unittest` from the standard library instead of importing it from `traits`.


rahulporuri:
thanks for the review @jvkersch

PR #447, opened 2019-06-06
Remove erroneous comment about label_rotation of LabelAxis.

jonathanrocher:
It is possible to specify a rotation angle for the labels of a `LabelAxis` to be any value: they don't have to be multiples of 90 and the comment is wrong. 

For example, this is what I am getting with 30:
Screen Shot 2019-06-06 at 5 14 00 PM


corranwebster:
LGTM, especially as the person who implemented the arbitrary rotation of labels ;)

jonathanrocher:
Thanks for looking at this quickly. And for the implementation!

Issue #456, opened 2019-06-25
BUG: DataBox Overlay Causing Component to Disappear


skailasa:
**Problem Description**

Resizing component containing DataBox failing due to bug in position update. It causes the component to 'disappear'

This is due to the fact that when updating the position, the first element of the transpose of `tmp` is taken which is the wrong dimension, tmp is a (1,2) array.meaning that first element of its transpose is just a (1,1) array leading to an error when attempting to unpack these values into xs and ys in `map_data`

```python
    # lines 144-152 in DataBox.py
    @on_trait_change('position,position_items')
    def _update_position(self):
        if self._updating:
            return
        tmp = self.component.map_data(self.position)
        if len(tmp.shape) == 2:
            tmp = tmp.T[0]
        self._data_position = tmp
        self.trait_property_changed("data_position", self._data_position)
```

**Reproduction Steps:**

  ```python
# Create traitsui view, with a component containing a DataBox

class Demo(HasTraits):
	
	plot = Instance(Plot)

	def _plot_default(self):
		data = ...
		plot = Plot(data)

		databox_overlay = DataBox(component=plot)
		move_tool = MoveTool(component=databox_overlay)
		databox_overlay.tools.append(move_tool)

		plot.overlays.append(databox_overlay)


	view = View(
	Item('plot')
	)


if __name__ == "__main__":
	Demo().configure_traits()
```

**Expected behavior:**

No errors when resizing plot component with DataBox overlay

**OS, Python version:** 
MacOS Python 3.60

PR #457, opened 2019-06-25
BUG: Fix DataBox resizing bug (#456)

skailasa:
Addresses issue #456 regarding disappearing image on DataBox resize, adds test.

### DIFF
--- a/chaco/overlays/tests/test_databox.py
+++ b/chaco/overlays/tests/test_databox.py
@@ -17,9 +17,9 @@ class TestDataBox(unittest.TestCase, UnittestTools):
         xs = np.arange(0, 5)
         ys = np.arange(0, 5)
 
+        # Add some fake 2D data for the box's component
         index = GridDataSource(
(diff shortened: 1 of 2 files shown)


chaco/overlays/tests/test_databox.py
```diff
@@ -0,0 +1,59 @@
...
+        """
+        Test that data position property is updated properly when the position
+            trait is changed.
+        """
+
+        # Check that data position property is changed when position is changed
+        with self.assertTraitChanges(self.databox, 'data_position') as result:
+            self.databox.position = [1, 1]
+
+            # Without moving the DataBox, data_position trait defaults
```
jvkersch: Everything from this point onwards can be moved out of the context manager.

skailasa: @jvkersch , I don't think I have write access.

jvkersch: Oops, I didn't think of that. Anyway, looks great, merging. Thanks for diagnosing this and providing the fix.

PR #460, opened 2019-07-08
REL: 4.8.0

jvkersch:
Release branch for 4.8.0

The actual release commit is 4437360

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -7,7 +7,7 @@ Release 4.8.0
 
 New features/Improvements
 
-* Refactor DataPrinter to add `_build_text_from_event` method (PR #450)
+* Refactor DataPrinter to add _build_text_from_event method (PR #450)
 * Simplify controlling text of ImageInspectorOverlay (PR #431)
 * 2D segment and text scatter plot types (PR #407)
 * Set up Windows testing under AppVeyor (PR #445)
@@ -37,7 +37,7 @@ Maintenance
 * Update unittest imports (PR #446)
 * Address DeprecationWarnings (PR #441)
 * Use unittest-style tests (PR #440)
-* Run testsuite on OSX and Python 3.6 (PR #442)
+* Run testsuite on mac OS and Python 3.6 (PR #442)
 * Use standard Sphinx extensions for trait documenting (PR #436)
 * Update Cython generated code to support Python 3.7 (PR #433)
 * Skip DataFramePlotData tests if Pandas not installed (PR #422)
@@ -50,7 +50,7 @@ Fixes
 
 * Ensure contiguous inputs to points_in_polygon (#409 & #410)
 * Handle multiple plots in LegendHighlighter (#403)
-* alias six.moves as sm (#401)
+* Alias six.moves as sm (#401)
 * Respect visibility in the LegendHighlighter (#402)
 * Fix use of itertools.chain() (#394)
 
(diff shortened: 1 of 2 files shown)


CHANGES.txt
```diff
@@ -2,12 +2,65 @@
 Chaco CHANGELOG
 ===============
 
+Release 4.8.0
+-------------
+
+New features/Improvements
+
+* Refactor DataPrinter to add `_build_text_from_event` method (PR #450)
```
mdickinson: Is this changelog thinking of itself as reST, Markdown, or neither? If reST, the backticks should be doubled.

CHANGES.txt
```diff
@@ -2,12 +2,65 @@
...
+* Fix up create_plot_snapshots script (PR #421)
+* Docstring for ArrayPlotData.{set_data,update_data} (PR #399)
+
+Maintenance
+
+* Update Travis CI configuration to be compatible with Ubuntu Xenial (PR #452)
+* Update unittest imports (PR #446)
+* Address DeprecationWarnings (PR #441)
+* Use unittest-style tests (PR #440)
+* Run testsuite on OSX and Python 3.6 (PR #442)
```
mdickinson: nitpick: macOS, not OSX

CHANGES.txt
```diff
@@ -2,12 +2,65 @@
...
+
+
+Release 4.7.2
+-------------
+
+Fixes
+
+* Ensure contiguous inputs to points_in_polygon (#409 & #410)
+* Handle multiple plots in LegendHighlighter (#403)
+* alias six.moves as sm (#401)
```
mdickinson: Nitpick: capitalization

CHANGES.txt
jvkersch: I removed the backticks altogether (the rest of the changelog doesn't have any code literals at all). We can make it into an official reST file later on.

jvkersch:
Thanks @mdickinson. I've fixed up the changelog.

mdickinson:
> That also doesn't help much unless enable does the same.

Not true: it does help, independently of Enable. The only other thing Enable needs is SWIG, and that's not something we can `pip install` anyway.

> with one test failure

Opened #468 to track this. It only fails under unittest, not under nose, and it looks like a shallow bug in the test itself, so definitely not a blocker for release purposes.

corranwebster:
@achabotl found and fixed a few circular import problems while working on #462. Options are:

- merge #462 as-is.  There is more work needed but it is currently in a stable state and further work could go into another PR if needed.
- wait for Alex and I to have time to finish #462 (docs need a conceptual reorganization and some de-duplication)
- cherry-pick the bugfixes into a separate PR and include that in the release.

achabotl:
The relevant commits are:

- a963fd32 FIX: Circular import in toolbar_plot
- 20ce8c5d FIX: Fix circular import in transform_color_mapper

And

- ef505c65 FIX: Add missing imports of `SegmentPlot` and `TextPlot` to `api.py`.

jvkersch:
@achabotl Would you be able to open a separate PR with those cherry-picked commits?

corranwebster:
I can do it this afternoon if @achabotl doesn't have time.

achabotl:
I pushed #469 for circular imports. and #470 for adding missing imports in `api.py`.

jvkersch:
Let's put this on hold until #469 is merged (#470 is already merged) and until we have a grip on #468.

PR #467, opened 2019-07-15
Declare build dependencies.

mdickinson:
This PR declares NumPy and Cython as build dependencies; without this, a `pip install` into an environment that lacks NumPy and/or Cython will fail.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -12,3 +12,4 @@ include CHANGES.txt
 include LICENSE.txt
 include image_LICENSE.txt
 include dev_requirements.txt
+include pyproject.toml
--- a/ci/edmtool.py
+++ b/ci/edmtool.py
@@ -130,7 +130,10 @@ def install(runtime, toolkit, environment):
         # Note that enable dependencies will be installed implicitly using pip
         ("edm run -e {environment} -- "
          "pip install git+https://git@github.com/enthought/enable.git"),
-        "edm run -e {environment} -- pip install . --no-deps",
+        # Use --no-build-isolation to avoid trying to build using a different
+        # NumPy version from the one currently installed.
+        ("edm run -e {environment} -- "
+         "pip install . --no-build-isolation --no-dependencies"),
     ]
     click.echo("Creating environment '{environment}'".format(**parameters))
     execute(commands, parameters)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,2 @@
+[build-system]
+requires = ["Cython", "numpy", "setuptools", "wheel"]


mdickinson:
Hmm. Two sources of unhappiness in the tests: missing fonts, and apparent NumPy binary incompatibilities. Examples of the latter:
```
ERROR: chaco.overlays.tests.test_databox (unittest.loader._FailedTest)
ImportError: Failed to import test module: chaco.overlays.tests.test_databox
  File "c:\users\appveyor\.edm\envs\chaco-test-3.6-null\lib\unittest\loader.py", line 428, in _find_test_path
    module = self._get_module_from_name(name)
  File "c:\users\appveyor\.edm\envs\chaco-test-3.6-null\lib\unittest\loader.py", line 369, in _get_module_from_name
    __import__(name)
  File "c:\users\appveyor\.edm\envs\chaco-test-3.6-null\lib\site-packages\chaco\overlays\tests\test_databox.py", line 6, in <module>
    from chaco.api import (
  File "c:\users\appveyor\.edm\envs\chaco-test-3.6-null\lib\site-packages\chaco\api.py", line 29, in <module>
    from .color_mapper import ColorMapper, ColorMapTemplate
  File "c:\users\appveyor\.edm\envs\chaco-test-3.6-null\lib\site-packages\chaco\color_mapper.py", line 19, in <module>
    from .speedups import map_colors, map_colors_uint8
  File "c:\users\appveyor\.edm\envs\chaco-test-3.6-null\lib\site-packages\chaco\speedups.py", line 8, in <module>
    from ._cython_speedups import *
  File "__init__.pxd", line 918, in init chaco._cython_speedups
ValueError: numpy.ufunc size changed, may indicate binary incompatibility. Expected 216 from C header, got 192 from PyObject
```

mdickinson:
So this is nasty. It looks as though the build environment is created without reference to existing packages, so even though there's a NumPy (version 1.15.x) already present in the environment when we do the `pip install .` for Chaco, the NumPy entry in the `pyproject.toml` means that the *latest* version of NumPy gets installed and used for building (and the same for Cython).

Replacing the `pip install .` in `edmtool.py` with `pip install -v .`, I see the following in the build output:
```
Using version 1.16.4 (newest of versions: 1.3.0, 1.4.1, 1.5.0, 1.5.1, 1.6.0, 1.6.1, 1.6.2, 1.7.0, 1.7.1, 1.7.2, 1.8.0, 1.8.1, 1.8.2, 1.9.0, 1.9.1, 1.9.2, 1.9.3, 1.10.0.post2, 1.10.1, 1.10.2, 1.10.4, 1.11.0, 1.11.1, 1.11.2, 1.11.3, 1.12.0, 1.12.1, 1.13.0, 1.13.1, 1.13.3, 1.14.0, 1.14.1, 1.14.2, 1.14.3, 1.14.4, 1.14.5, 1.14.6, 1.15.0, 1.15.1, 1.15.2, 1.15.3, 1.15.4, 1.16.0, 1.16.1, 1.16.2, 1.16.3, 1.16.4)
```

prabhuramachandran:
@mdickinson -- I have found pyproject.toml to be more trouble than it is worth for exactly these sort of issues with dependencies and version incompatibilities.  Enough issues to simply not use it at all, in multiple other projects with numpy dependencies.  I am not sure it has improved but this was really frustrating for me.

mdickinson:
@prabhuramachandran Right, possibly this PR should just be closed. I don't think I understand how this is *supposed* to be working (at Python packaging level). It's presumably fine in a situation where (a) you're installing everything from PyPI, and (b) you can use the latest versions of everything, but it seems flawed in this sort of case. I don't think our difficulties here are just down to using a mixture of EDM and PyPI; it seems to be a more fundamental problem than that.

prabhuramachandran:
@mdickinson -- yes this is not just about edm and pypi.  Many projects with numpy dependencies seem to run into this. A useful discussion was here: https://github.com/numpy/numpy/issues/5888 and it does look like there are some possible ways out of this by choosing a lower numpy version.

mdickinson:
Closing this.

PR #471, opened 2019-07-18
Fix a test that fails with NumPy 1.16.x.

### LINKED ISSUE #468: test_selection_no_warning fails under unittest
When running the test suite against an installed Chaco with `python -m unittest discover -v chaco` (from a directory far away from the GitHub checkout), I see one test failure:

```
FAIL: test_selection_no_warning (chaco.tools.tests.test_range_selection.RangeSelectionTestCase)
  File "/Users/mdickinson/Desktop/chacotest/lib/python3.7/site-packages/chaco/tools/tests/test_range_selection.py", line 70, in test_selection_no_warning
    self.assertEqual(w, [])
AssertionError: Lists differ: [<warnings.WarningMessage object at 0x11eea8f90>] != []

First list contains 1 additional elements.
First extra element 0:
<warnings.WarningMessage object at 0x11eea8f90>

- [<warnings.WarningMessage object at 0x11eea8f90>]
+ []
```

The test passes under nose.

Seen under both Python 3.6 and Python 3.7 on macOS 10.14.5 with Chaco installed from the soon-to-be 4.8.0 commit (44373605d248602dbdaf51547f65e5a90fec00e7).

mdickinson:
`test_selection_no_warning` is checking that Chaco itself doesn't raise any warnings on selection assignment. Previously, we were filtering out a known NumPy warning; that filter doesn't work with NumPy 1.16.x because the warning message changed.

This PR updates the filter to filter out all warning messages from packages other than Chaco.

Fixes #468.

### DIFF
--- a/chaco/tools/tests/test_range_selection.py
+++ b/chaco/tools/tests/test_range_selection.py
@@ -61,9 +61,10 @@ class RangeSelectionTestCase(EnableTestAssistant, unittest.TestCase):
         renderer = plot.plot(('x', 'y'))[0]
         tool = RangeSelection(renderer)
         with warnings.catch_warnings(record=True) as w:
-            # Ignore warnings coming from Traits
+            # Ignore warnings coming from any package other than Chaco
             warnings.filterwarnings(
-                "ignore", 'elementwise == comparison failed'
+                "ignore",
+                module="(?!chaco)",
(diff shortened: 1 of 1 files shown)


jvkersch:
LGTM!

PR #477, opened 2019-08-05
DOC: 4.8.0

jvkersch:
Updates the Chaco documentation. This is a large PR, likely because we switched in the 4.8 release from using Chaco's own Sphinx extensions to using trait-documenter and Napoleon.


jvkersch:
@corranwebster @mdickinson Could you give this a review? 

I moved the old (4.6) docs out of the way, and added the new (4.8) docs at the root level.

mdickinson:
There are some extras here that you probably don't want/need:

- there's a dist/ directory containing chaco-4.8.0.tar.gz; presumably you're not intending to include that in the doc distribution?
- the "chaco" and "chaco.egg-info" directories should be removed
- .gitignore isn't needed (but is mostly harmless)
- .buildinfo isn't needed (ditto)

Otherwise, looks okay.

It would be good to update to the Enthought Sphinx Theme at some point; we're converging on that theme for all other ETS projects. It's easier than it used to be to do that - there's an up-to-date egg in the Enthought buildsystem and a corresponding sdist on PyPI.

mdickinson:
BTW, where did the ".nojenkins" file come from, and what's it for? Do you have a relevant documentation link? (Wondering whether I need to use it on other ETS projects.)

mdickinson:
Ah, @pberkes added it as part of [this](https://github.com/enthought/chaco/commit/921bd451ed7360aae6cf97043dfced5602a27a14) commit.

I don't think we need it.

jvkersch:
@mdickinson Where exactly do you see the `dist`, `chaco` and `chaco.egg-info` files? If they're in the root directory, they may be artifacts of a previous chaco checkout.  Here's how my checkout looks:
```
~/src/enthought/chaco [gh-pages-4.8.0 ↑·1|⚑ 1] 
17:37 $ ls -a
.                           _images                     scales.html
..                          index.html                  search.html
4.6                         modules_and_classes.html    searchindex.js
api                         .nojekyll                   _sources
architecture_overview.html  objects.inv                 _static
_downloads                  programmers_reference.html  tech_notes.html
.git                        quickstart.html             user_manual
```

I've deleted the other files that you mentioned.

jvkersch:
I've opened an issue for switching to the Enthought Sphinx theme: https://github.com/enthought/chaco/issues/479

mdickinson:
> they may be artifacts of a previous chaco checkout

Indeed. :flushed:

Sorry.

jvkersch:
Thanks @mdickinson !

PR #482, opened 2019-08-25
Add rectangular selection tool

notmatthancock:
This adds a subclass, `RectangularSelection`, of the `LassoSelection` tool that allows selection via a rectangular bounding box. This is a much quicker user interaction for situations where the selection region does not need to be as precise as the lasso tool allows or when the data being selected lie on a regular grid.

This also adds (yet another!) scatter selection demo script, which augments the lasso selection demo:

rect-select


corranwebster:
This looks good, and nicely done - a neat approach that I wouldn't have thought of to get this effect.

I know that the superclass doesn't have any tests, but any chance for tests for this?

notmatthancock:
> I know that the superclass doesn't have any tests, but any chance for tests for this?

Sure, I added a test to check for selecting the middle two of four co-linear points.

corranwebster:
Thanks.  LGTM.  I'll leave it up to @jvkersch to manage the merge unless there is some urgency.

PR #487, opened 2019-10-30
Fix attempts to get the .shape attribute of a list or tuple.

mdickinson:
This PR fixes an obvious bug in the `LinearMapper` code, where we attempt to look up the `shape` attribute of a list or a tuple.

Code ref: https://github.com/enthought/chaco/blob/e0e4f7d8bc5551191ad65295fe97f55190e87a9c/chaco/linear_mapper.py#L47-L50

### DIFF
--- a/chaco/horizon_plot.py
+++ b/chaco/horizon_plot.py
@@ -1,6 +1,6 @@
 from __future__ import with_statement
 
-from numpy import array, transpose, ndarray, empty
+from numpy import array, float64, full_like, ndarray, transpose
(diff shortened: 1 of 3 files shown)


chaco/horizon_plot.py
```diff
@@ -82,7 +80,7 @@ def _render(self, gc, points):
         # Get color bands
         bands = array(self.color_mapper._get_color_bands())
 
-        with gc: 
+        with gc:
```
mdickinson: My editor made these fixes, by auto-trimming trailing whitespace. I can revert them if that's preferred.

PR #494, opened 2019-11-26
Add a new renderer for LOD images 

xiaoyu-wu:
A new pair of image source and renderer are added to visualize an image larger than memory size and has multiple level of details (LODs). Images are rendered with increasing resolution so that the UI remain responsive during scrolling.
lod_viewer_example


chaco/lod_image_plot.py
```diff
@@ -0,0 +1,322 @@
...
+    KIVA_INTERP_QUALITY = {"nearest": InterpolationQuality.none,
+                           "bilinear": InterpolationQuality.low,
+                           "bicubic": InterpolationQuality.high}
+try:
+    from encore.concurrent.futures.serializer import Serializer
+    from encore.concurrent.futures.enhanced_thread_pool_executor import \
+        EnhancedThreadPoolExecutor
+except ImportError:
+    import sys
+    sys.exit('encore must be installed to use LODImagePlot')
```
jwiggins: You shouldn't do this if you're going to import from this module in `chaco.api`.

(disregard if encore is already a hard requirement)

xiaoyu-wu: I realized that after seeing the CI failed. `encore` is not a hard requirement ATM.

corranwebster:
I'd like to think a bit about how to generalize this and possibly combine with the (largely stubbed out) `downsample` methods.  The core ideas are good.

chaco/lod_image_plot.py
xiaoyu-wu: Do you know any better place to import this module from? Are there anything like `chaco.experimental.api`?

jwiggins: No. I think the answer it to just leave it out and force users to import it directly.

chaco/lod_image_plot.py
```diff
@@ -0,0 +1,322 @@
...
+    from kiva.quartz.ABCGI import InterpolationQuality
+except ImportError:
+    pass
+else:
+    KIVA_INTERP_QUALITY = {"nearest": InterpolationQuality.none,
+                           "bilinear": InterpolationQuality.low,
+                           "bicubic": InterpolationQuality.high}
+try:
+    from encore.concurrent.futures.serializer import Serializer
+    from encore.concurrent.futures.enhanced_thread_pool_executor import \
```
corranwebster: I think that using Traits Futures is probably preferable: https://github.com/enthought/traits-futures . The `TraitsExecutor` in that library is probably a drop-in replacement (or close to it).

corranwebster:
Playing around with the example, I'm noticing a few issues with the UX which would be nice to fix.

The first is that it is very jittery when moving, as it renders the first few LoD before getting the bounds change event - it would be nice to have some mechanism to estimate the initial LoD based on the render time we have (rough guideline is that you have 50-100ms for the first render) - since we anticipate that getting the data is the slow operation, we might be able to step through the first few LoD without rendering if we get the data fast enough.

The second is that there may be an issue with bounds estimation, in that there is a jittery white area  in the direction that is being scrolled.  Either we're not estimating the slices correctly and so underpainting, or the area we are painting on is getting out of sync with the movement.

Thirdly, if I replace the scrollbars with the more usual pan and zoom tools I get distortion of the image while panning.  This would also point to the slicing not being done correctly.

As much as possible with this we'd like the UX for LoD to be as unobtrusive as possible.

chaco/lod_image_plot.py
xiaoyu-wu: Moved as suggested.

chaco/abstract_data_source.py
```diff
@@ -56,7 +56,7 @@ class AbstractDataSource(HasTraits):
     # Abstract methods
     #------------------------------------------------------------------------
 
-    def get_data(self):
+    def get_data(self, lod=None):
```
corranwebster: Should update the docstring below to describe the new optional argument

chaco/image_plot.py
```diff
@@ -43,7 +44,7 @@
 KIVA_DEPTH_MAP = {3: "rgb24", 4: "rgba32"}
 
 
-class ImagePlot(Base2DPlot):
+class ImagePlot(Base2DPlot, Handler):
```
corranwebster: This shouldn't inherit from `Handler`.  This creates a problem with the life-cycle of the executor - we may need Enable to send a new type of event to indicate that the plot is being closed.

chaco/abstract_data_source.py
xiaoyu-wu: Updated.

chaco/image_plot.py
xiaoyu-wu: How about overriding the `cleanup` method of `Component` and adding codes that stop the executor there?

corranwebster:
I think the main remaining concern I have is around the life-cycle of the executor.  I had some discussion with Mark Dickinson about this this morning and his feeling was that the executor should be controlled by the application and passed to the plot, which at a minimum means that the executor should be a public trait not a private one, but also probably means that we shouldn't try to clean it up _unless_ we know that the plot created it as a default rather than user code passing it in (noting that on trait change handlers have a distressing tendency to create defaults for the old value only to immediately throw it away).

Which brings up the second part of the concern, which is that the `cleanup` method isn't guaranteed to be called: in particular if the plot is removed from its container before the window is closed, then the `cleanup` method never gets called.  This is probably a bug in Enable - `cleanup` should probably be called as part of `Component.remove` as well.  Fixing the Enable bug is outside the scope of the PR, but we should be aware of it.

Other than these concerns, the PR is coming together nicely.

xiaoyu-wu:
> I think the main remaining concern I have is around the life-cycle of the executor. 

Thanks for your thoughts on this. I agree that the traits executor is better a public trait and passed in during instantiate. (In fact, the very original commit was using a passed-in job scheduler.) 

Two possible solutions are:
1) Enforce that a trait executor should be passed in during instantiation of the `ImagePlot` object if `use_downsampling` is True. In this case, creating / destroying of the executor should be handled explicitly by the user.
2) Expect a passed-in executor but also create one if not provided. In this case, we need a flag to help remember if the executor is created by this plot and be responsible for its destruction, which I believe introduced too much complexity that is not strongly related to the core functions of this class.

I prefer the 1st solution. Would you agree? @corranwebster

mdickinson:
I also prefer the 1st solution, requiring an executor. That always leaves open the possibility of making it optional later on (i.e., moving from solution 1 to solution 2) if there's demand for that. Going the other way (implementing solution 2, realising that it's problematic, and then removing functionality) would be harder.

chaco/image_plot.py
```diff
@@ -372,3 +421,18 @@ def _array_bounds_from_screen_rect(self, image_rect):
         row_max = min(row_max, array_height)
```
siddhantwahal, line 421: An edge case that isn't handled here occurs when `col_min` is greater than `array_width` or `col_max` is less than `0`. These only occur when the image disappears from the plot, e.g., panning the image to the right as much as possible and then zooming in leads to  negative `x_max` and `col_max`. 

While `x_min` and `col_min` are also negative in this example, `col_min` is explicitly clipped to 0, whereas `col_max` isn't, leading to `col_max - col_min < 0`.

When these bounds are used to compute the LOD in `_calculate_necessary_lod`:
https://github.com/enthought/chaco/blob/e41cfd375d611d3438328ced67fbe9a4c3182c39/chaco/image_plot.py#L432-L438
 no LOD will satisfy `col_max - col_min >= screen_rect[2]`, and so the necessary LOD is set to 0. This causes the highest resolution image to be unnecessarily cached, potentially crashing the program if the image is too large to fit in memory (actually, only the slice `[:, :-abs(col_max)]` is cached, but that can still be large).

The fix is simple:
```diff
-        col_min = max(col_min, 0)
-        col_max = min(col_max, array_width)
-        row_min = max(row_min, 0)
-        row_max = min(row_max, array_height)
+        col_min, col_max = np.clip([col_min, col_max], 0, array_width)
+        row_min, row_max = np.clip([row_min, row_max], 0, array_height)
```

corranwebster:
FWIW - some other Enthought teams have reported success using traits-futures but just using it to directly update the plot data as LoD information comes in, without needing to integrate traits futures into the plotting code.

PR #497, opened 2019-12-23
Show contents of demo folder in TreeView

### LINKED ISSUE #496: Demo application doesn't show all demos in tree view
**Problem Description**

`examples/demo/demo.py` script doesn't show contents of `examples/demo` directory, only subfolders.

**Reproduction Steps:**

With Chaco installed, cd to `examples/demo` and run `python demo.py`

**Expected behavior:**

Files directly in the `examples/demo` subfolder should be shown in the tree view.

**OS, Python version:** All, all.

Note that this may require fixing in TraitsUI, which is where the core of the `demo.py` script is defined.

midhun-pm:
Fixes #496 
Just needed to add a flag and turn the demo folder into a module

### DIFF
--- a/examples/demo/demo.py
+++ b/examples/demo/demo.py
@@ -21,6 +21,7 @@
 from traitsui.extras.demo import demo
 
 demo(
+    use_files=True,
     title='Chaco Demos'
 )
 


corranwebster:
LGTM

PR #498, opened 2019-12-26
Fix stacked bar plot example, by using the correct starting_value

### LINKED ISSUE #444: bar_plot_stacked example unfinished?
**Problem Description**
The example in https://github.com/enthought/chaco/blob/master/examples/demo/basic/bar_plot_stacked.py
doesn't do any stacking.

**Expected behavior:**
I the bars were really stacked, I would expect the sum of all bars to reach (10+5+2) * array([1,2,3,4,5]) (the sum of all values) respectively. Instead, I am getting the following:
Screen Shot 2019-05-31 at 10 19 28 AM

Looking at the code, it doesn't use the bar plot's `starting_value` as expected, so the demo doesn't even seem to try to do the right thing.

midhun-pm:
Fixes #444

### DIFF
--- a/examples/demo/basic/bar_plot_stacked.py
+++ b/examples/demo/basic/bar_plot_stacked.py
@@ -11,7 +11,7 @@ from traits.api import HasTraits, Instance
 from traitsui.api import UItem, View
 
 # Chaco imports
-from chaco.api import LabelAxis, Plot, ArrayPlotData
+from chaco.api import LabelAxis, Plot, ArrayPlotData, ArrayDataSource
 
 class PlotExample(HasTraits):
     plot = Instance(Plot)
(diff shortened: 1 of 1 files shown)


examples/demo/basic/bar_plot_stacked.py
```diff
@@ -47,7 +47,7 @@ def __init__(self, index, series_a, series_b, series_c, **kw):
 
 
 index = numpy.array([1,2,3,4,5])
-demo = PlotExample(index, index*10, index*5, index*2)
+demo = PlotExample(index, index*10, index*15, index*17)
```
shoeb-github: Is it not possible to do this without changing the input data?

midhun-pm: Docstring for `starting_value` says the following:
```   
    #: The data source to use as "starting" values for bars (along value axis).
    #: For instance, if the values are [10, 20] and starting_value
    #: is [3, 7], BarPlot will plot two bars, one  between 3 and 10, and
    #: one between 7 and 20
```
I think `BarPlot` is designed not to change the value when stacking, so the user may have to do it explicitly.

corranwebster: You could reverse the input data instead.  The original version of this relied on the rendering order (the tallest bar was drawn first, etc.).

Even better might be to sum the data as part of the process of creating the plots (ie. barplot 1 is from 0 to `series_a`, barplot 2 is from `series_a` to `series_a + series_b`, etc.)

midhun-pm: Agreed, that's better, made the change.

Issue #359, opened 2017-06-27
[Examples] contour_plot demo has issues


rahulporuri:
specifically, `examples/demo/basic/contour_plot.py`.
Running the script works fine but the plot generated is what has a few issues.

1. There is a white box in the contour plot, which I assume is a legend. See below.

![screen shot 2017-06-27 at 9 19 14 am](https://user-images.githubusercontent.com/1926457/27570209-00540b88-5b1a-11e7-95ec-b9644f43e5eb.png)

2. _I don't know if this is the expected behavior but_ Zooming in and out using the scroll wheel doesn't change the plot, unless I pan/move the plot.

Reproduced on both Py2.7.13 and Py3.6.0 env. OSX.

corranwebster:
1) The white box is demonstrating the (new) capability to mask regions of the plot (see #326)

2) Did the plot have focus when you first used the scroll wheel? Ie. I would expect that you may not get scroll wheel zoom until you have at least clicked on the plot or the window containing it.

rahulporuri:
> Did the plot have focus when you first used the scroll wheel? Ie. I would expect that you may not get scroll wheel zoom until you have at least clicked on the plot or the window containing it.

You're right. I wasn't able to get the scroll wheel zoom to work until I clicked on the plot. I don't quite understand what's going here so i'm sorry if im missing something trivial/straight-forward.

corranwebster:
If you haven't run into the concept of "focus" in UIs before, it's just the notion of a widget being the target of events which aren't specifically tied to a point on the screen - most notably being which widget should get keystrokes when the user types, but the scroll wheel potentially being another (possibly depending on the oS you are running on, as well).  And if you have run into the concept I apologise for explaining it 😉 

So I think what is happening is that when you click on the plot, it (or at least the QWidget it is embedded in) gets focus and starts being able to get scroll wheel events. This may be unexpected in a modern UI, because browser based UIs tend to be a bit more lax about the idea of focus.

Still, it might be worth looking into the Enable event code to check that we're not doing something dumb with the scroll events.

jvkersch:
@rahulporuri @corranwebster Is there something to do here, or can this issue be closed?

jvkersch:
@rahulporuri I'll close this issue since there's nothing actionable anymore, but feel free to reopen if I've overlooked anything.

PR #506, opened 2020-01-02
Fix: Zoom not working with contour plots

### LINKED ISSUE #466: Zoom broken in some demos
**Problem Description**
This bug affects the following demos:
* `shell/contour.py`
* `shell/contourf.py`
* `basic/contour_plot.py`

Mouse-scroll doesn't zoom the plot until the user pans the plot. After panning, the plot is really zoomed out.

**Reproduction Steps:**
Mouse-scroll, then left-drag.

```python
python shell/contour.py
```

**Expected behavior:**
Plot should zoom immediately on mouse-scroll.

**OS, Python version:**
MacOSX 10.14
Python3.6

shoeb-github:
closes #466. Also reported in #359.

Fixed by requesting redraw when index mapper is updated in contour plots.

### DIFF
--- a/chaco/base_contour_plot.py
+++ b/chaco/base_contour_plot.py
@@ -142,13 +142,13 @@ class BaseContourPlot(Base2DPlot):
         # If the index data has changed, the reset the levels cache (which
         # also triggers all the other caches to reset).
         self._level_cache_valid = False
-        self.invalidate_draw()
+        self.invalidate_and_redraw()
 
     def _value_data_changed_fired(self):
         # If the index data has changed, the reset the levels cache (which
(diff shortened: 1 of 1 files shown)


chaco/base_contour_plot.py
```diff
@@ -153,6 +153,7 @@ def _value_data_changed_fired(self):
     def _index_mapper_changed_fired(self):
         # If the index mapper has changed, then we need to redraw
         self.invalidate_draw()
+        self.request_redraw()
```
midhun-pm: Seems to work. Not sure what `self.invalidate_draw()` does as the zoom works even without it.

PR #509, opened 2020-01-16
Import from traits.api instead of from traits.has_traits

mdickinson:
For safety against future Traits internal refactors, import from traits.api whenever possible.


jonathanrocher:
LGTM.

PR #513, opened 2020-02-05
Fix and clean up multiaxis.py example

jonathanrocher:
- Aligned j0, j2 and j3's `value_mapper` so curves are correctly placed. To some extent, that changes a bit the purpose of the example to be more like the kind of thing that's scientifically useful (as opposed to disconnecting 4 plot's y mappers, but only displaying 2 axis).
- Added axis titles to clarify where to read y-values.
- Added a zoom tool because it is tricky to do and universally needed.
- pep8 fixes.


jonathanrocher:
Thanks for the review! Merging...

PR #514, opened 2020-03-03
Feature: support renderer traits

jonathanrocher:
These factory functions are used all over the place in the demos. This PR adds:
1. docstrings for the factory functions.
2. unit tests for the factory functions.
3. a `**renderer_traits` argument to support passing more renderer traits through the factory.

Even though my eyes bleed, I am refraining from adding pep8 fixes to keep the PR clean.

### DIFF
--- a/chaco/tests/test_plot_factory.py
+++ b/chaco/tests/test_plot_factory.py
@@ -12,7 +12,7 @@ x = np.array([1, 2, 3, 4])
 y = np.array([1, 2, 3, 4])
 
 
-class BaseTestRenderer(object):
+class BaseTestRenderer:
     def test_create_renderer_default(self):
         renderer = self.factory((x, y))
         self.assertIsInstance(renderer, self.renderer_klass)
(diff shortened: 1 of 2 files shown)


chaco/tests/test_plot_factory.py
```diff
@@ -0,0 +1,68 @@
...
+    create_polar_plot, create_scatter_plot
+from chaco.api import BarPlot, LinePlot, PlotAxis, PlotGrid, ScatterPlot
+from chaco.polar_line_renderer import PolarLineRenderer
+
+x = np.array([1, 2, 3, 4])
+
+y = np.array([1, 2, 3, 4])
+
+
+class BaseTestRenderer(object):
```
jvkersch: You can drop the `object` here -- we're no longer explicitly supporting Python 2.

jonathanrocher:
Thanks for the review @jvkersch ! Will merge once CI is happy.

PR #517, opened 2020-05-07
ENH: Setup online redirect and proper index page for docset

achabotl:
Adds doc2dash flags to:

- set the online redirect url (`-u`) so one can go from the local docsets to the online ones.
- Set the index page (`-I`) to show `index.html`, which is what is showed online, instead of the index of all the names.


jvkersch:
LGTM! I verified that the docsets build and I can load them into Zeal, with the desired changes.

The build fails because we are still testing on Python 2.7... I will rip this out, but this PR can be merged in the meantime.

PR #524, opened 2020-07-13
Remove Python 2.7, 3.5 testing, add PyQt 5

jvkersch:
as it says on the tin.

Note: this means that we effectively don't test with wxpython and pyside any more. 

@rahulporuri Do you know of any other toolkits that are supported under Python 3.6?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,12 +18,6 @@ env:
 
 matrix:
   include:
-  - env: RUNTIME=2.7 TOOLKITS="null pyqt pyside wx"
-  - os : osx
-    env: RUNTIME=2.7 TOOLKITS="null pyqt pyside wx"
-  - env: RUNTIME=3.5 TOOLKITS="null pyqt"
-  - os : osx
-    env: RUNTIME=3.5 TOOLKITS="null pyqt"
   - env: RUNTIME=3.6 TOOLKITS="null pyqt"
   - os : osx
     env: RUNTIME=3.6 TOOLKITS="null pyqt"
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ addons:
     - ccache
     - cmake
     - swig
+    - libxkbcommon-x11-0
     - libglu1-mesa-dev
     - libxcb-icccm4
     - libxcb-image0
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,9 +26,9 @@ env:
 
 matrix:
   include:
-  - env: RUNTIME=3.6 TOOLKITS="null pyqt pyqt5"
+  - env: RUNTIME=3.6 TOOLKITS="null pyqt pyqt5 pyside2"
   - os : osx
-    env: RUNTIME=3.6 TOOLKITS="null pyqt pyqt5"
+    env: RUNTIME=3.6 TOOLKITS="null pyqt pyqt5 pyside2"
   fast_finish: true
 
 branches:
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,9 +18,9 @@ env:
 
 matrix:
   include:
-  - env: RUNTIME=3.6 TOOLKITS="null pyqt"
+  - env: RUNTIME=3.6 TOOLKITS="null pyqt pyqt5"
   - os : osx
-    env: RUNTIME=3.6 TOOLKITS="null pyqt"
+    env: RUNTIME=3.6 TOOLKITS="null pyqt pyqt5"
   fast_finish: true
 
 branches:
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,11 +10,18 @@ addons:
     - cmake
     - swig
     - libglu1-mesa-dev
-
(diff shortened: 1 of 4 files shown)


corranwebster:
You should be able to get things to run with pip-installed `pyside2`, with a bit of luck.  Have a look at the TraitsUI and Pyface `etstools.py` to see how we do it there.

rahulporuri:
> @rahulporuri Do you know of any other toolkits that are supported under Python 3.6?

Like corran already mentioned, we could test against `pyside2`.

ci/edmtool.py
```diff
@@ -94,13 +89,15 @@
 extra_dependencies = {
     'pyside': {'pyside'},
     'pyqt': {'pyqt'},
+    'pyqt5': {'pyqt5'},
     'wx': {'wxpython'},
     'null': set()
 }
 
 environment_vars = {
     'pyside': {'ETS_TOOLKIT': 'qt4', 'QT_API': 'pyside'},
```
rahulporuri, line 98: we can probably remove this (or maybe we will need to update it if we start testing against `pyside2`)

ci/edmtool.py
```diff
@@ -94,13 +89,15 @@
 extra_dependencies = {
     'pyside': {'pyside'},
```
rahulporuri, line 90: same comment here about either removing this or updating it if we add `pyside2`

rahulporuri:
ack. travis linux choked when it tried running tests against pyqt5 environment with the following message

```
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.

This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: minimal, offscreen, vnc, xcb.
```

We will probably need to make the following changes to the CI setup/environment - https://github.com/enthought/envisage/pull/287

.travis.yml
```diff
@@ -10,23 +10,24 @@ addons:
     - cmake
     - swig
     - libglu1-mesa-dev
-
+    - libxcb-icccm4
```
rahulporuri, line 13: looks like you might also need `    - libxkbcommon-x11-0`

rahulporuri:
travis linux still fails - with the better error message

```
QLibraryPrivate::loadPlugin failed on "/home/travis/.edm/envs/chaco-test-3.6-pyqt5/plugins/platforms/libqxcb.so" : "Cannot load library /home/travis/.edm/envs/chaco-test-3.6-pyqt5/plugins/platforms/libqxcb.so: (libxcb-xkb.so.1: cannot open shared object file: No such file or directory)"
```

.travis.yml
```diff
@@ -9,24 +9,26 @@ addons:
...
 matrix:
   include:
-  - env: RUNTIME=2.7 TOOLKITS="null pyqt pyside wx"
-  - os : osx
-    env: RUNTIME=2.7 TOOLKITS="null pyqt pyside wx"
-  - env: RUNTIME=3.5 TOOLKITS="null pyqt"
-  - os : osx
-    env: RUNTIME=3.5 TOOLKITS="null pyqt"
-  - env: RUNTIME=3.6 TOOLKITS="null pyqt"
+  - env: RUNTIME=3.6 TOOLKITS="null pyqt pyqt5"
```
rahulporuri, line 29: do we want to test pyside2 on osx/linux like we are on win?

ci/edmtool.py
```diff
@@ -43,8 +43,8 @@
 using::
     python edmtool.py test_all
 
-Currently supported runtime values are ``2.7`` and ``3.5``, and currently
-supported toolkits are ``null``, ``pyqt``, ``pyside``, and ``wx``.  Not all
+Currently supported runtime values are ``3.6``, and currently
+supported toolkits are ``null``, ``pyqt``, and ``pyqt5``.  Not all
```
rahulporuri, line 47: we need to add `pyside2` here.

.travis.yml
jvkersch: yes!

jvkersch:
@rahulporuri Thanks for the quick turnaround on reviewing. I think this now works, care to give it a final look? I had to make one non-trivial change for pyside2: under pyside2, instantiating a `QFrame` with a keyword argument `size` does not work. The workaround is to omit the `size` argument, and to resize the window afterwards. You can see the original failure with the following code snippet:

```python
from pyface.qt import QtCore, QtGui
app = QtGui.QApplication([])
size = QtCore.QSize(10, 10)
f = QtGui.QFrame(size=size)  # complains about missing setSize method
```

The commit history is cluttered, so we should merge this with squash-merge.

jvkersch:
Ha, you beat me to it :)

rahulporuri:
yep yep I wasn't sure why `QFrame.setSize` doesn't exist on PySide2 but exists on PyQt5. I didn't look closely enough at the PyQt5 docs but the PySide2 docs for `QFrame` didn't mention any `setSize`

PR #532, opened 2020-10-26
Add source click option used on cron jobs to install ets packages from source

### LINKED ISSUE #522: Setup cron job to test against dependencies from source
Travis (and possibly appveyor too) allow the creation of cron jobs. Some of the ETS packages now have cron jobs setup to test against dependencies which are installed from source - instead of testing against the released versions.

See examples : 

- https://github.com/enthought/envisage/pull/162
- https://github.com/enthought/traitsui/pull/914
- https://github.com/enthought/pyface/pull/549

aaronayres35:
fixes #522

This PR adds a `--source` click option and uses this option on cron jobs to install other ets packages from source.
Note that previously enable was installed from source always and that was how traits, traitsui and pyface got installed.  Following Marks comment here: https://github.com/enthought/envisage/pull/325#issuecomment-713410126, I added all of these to `__requires__` in `__init__.py`.  Additionally, I made it so that by default enable is not installed from source, but it will be when the `--source` flag is used.  

This follows what is done in other ets projects (e.g https://github.com/enthought/envisage/blob/8245931f7e3dfcb6513bfb7dc2ae9b10ca6f89a1/etstool.py#L255-L273)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ env:
     - INSTALL_EDM_VERSION=2.0.0
       PYTHONUNBUFFERED="1"
       QT_DEBUG_PLUGINS="1"
+    - TRAVIS_EVENT_TYPE='cron'
 
 matrix:
   include:
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,7 +53,13 @@ before_install:
 
   - edm install -y wheel click coverage
 install:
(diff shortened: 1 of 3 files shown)


rahulporuri (approved):
LGTM. Can you do the set `TRAVIS_EVENT_TYPE` env var in CI like with the other PRs to exercise the cron job code path in CI? I tested this locally and it works as expected.

rahulporuri:
I'm a little surprised at the fact that the CI passed without needing enable from master.

PR #538, opened 2020-10-28
Use PySide2 from EDM instead of PyPI and Use the latest EDM version (3.0.1)

### LINKED ISSUE #527: Use the latest EDM version (3.0.1)
A new major version of EDM is available and it would be good to update to the latest version - https://assets.enthought.com/downloads/edm/. This should be as simple as updating the environment variables in travis and appveyor config files.

https://github.com/enthought/chaco/blob/3de7780561fa29e79c887432d3ce408ea82d1614/.travis.yml#L23
https://github.com/enthought/chaco/blob/3de7780561fa29e79c887432d3ce408ea82d1614/appveyor.yml#L8

### LINKED ISSUE #536: Use PySide2 from EDM instead of PyPI
Now that PySide2 is available via EDM, we should try installing it using edm instead of using pip when setting up the development environment - https://github.com/enthought/chaco/blob/82ac23214cdd5ba16c81351b41c102b62563eb30/ci/edmtool.py#L150-L154.

aaronayres35:
fixes #536 

Also fixes #527 

In this PR we try to install pyside2 from edm instead of pip.  In order to do so, we need to use the latest version of edm in CI, so those version numbers are bumped as well.

In retrospect these should've been two separate PRs (shame!) and if a reviewer would prefer I am happy to go back and split them.  Initially I was just trying to fix 536 without realizing it required a fix for 527.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ addons:
     
 env:
   global:
-    - INSTALL_EDM_VERSION=2.0.0
+    - INSTALL_EDM_VERSION=3.0.1
       PYTHONUNBUFFERED="1"
       QT_DEBUG_PLUGINS="1"
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -5,7 +5,7 @@ environment:
 
   global:
     PYTHONUNBUFFERED: "1"
-    INSTALL_EDM_VERSION: "2.0.0"
+    INSTALL_EDM_VERSION: "3.0.1"
 
   matrix:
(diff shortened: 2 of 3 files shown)


aaronayres35:
Appveyor is failing with 
```
Error: no such option: --add-repository
Command exited with code 1
```
Im guessing this is because the edm version is 2.0.0 in the appveyor.yml file.

rahulporuri (approved):
LGTM. I went through the CI logs to make sure that there was no funny business with pyside2.

PR #540, opened 2020-11-12
Handle index=None in ScatterInspector

### LINKED ISSUE #26: ScatterInspector does not deselect all on index == None
I've been playing around with the scattertoggle.py example. Maybe I'm missing something, but the ScatterInspector seems to fail to deselect all points when clicking on whitespace on the plot (that is, index == None). Something along the lines of this seems to fix it.

``` python
class PatchedScatterInspector(ScatterInspector):
        def _deselect(self, index=None):
            plot = self.component
            if index:
                super(PatchedScatterInspector, self)._deselect(index)
            else:
                for name in ('index', 'value'):
                    if not hasattr(plot, name):
                        continue
                    md = getattr(plot, name).metadata
                    md[self.selection_metadata_name] = []
                    getattr(plot, name).metadata_changed = True
            return
```

Cheers,

-A

notmatthancock:
Closes #26

### DIFF
--- a/chaco/tools/scatter_inspector.py
+++ b/chaco/tools/scatter_inspector.py
@@ -125,6 +125,11 @@ class ScatterInspector(SelectTool):
                 # Only issue 1 event:
                 if name == 'index':
                     self.inspector_event = insp_event
+            elif index is None:
+                md[self.selection_metadata_name] = []
+                # Only issue 1 event:
+                if name == 'index':
+                    self.inspector_event = insp_event
         return
 
     def _select(self, index, append=True):


corranwebster (approved):
LGTM

rahulporuri:
Thanks for the fix @notmatthancock . Do you need a bugfix release of chaco to make use of this in the near future (within a month)?

notmatthancock:
> Do you need a bugfix release of chaco to make use of this in the near future (within a month)?

Sure. We're working around it, but we could remove our workaround code with a release.

PR #554, opened 2021-01-06
Remove use of six and __future__ imports

aaronayres35:
Part of #483 

This PR removes all use of `six` and also removes all `__future__` imports (only imports were for `with_statement`, `print_function`, and `absolute_import`).

It also drive by removes a few lines of commented out code

### DIFF
--- a/chaco/axis_view.py
+++ b/chaco/axis_view.py
@@ -1,7 +1,4 @@
 """ Defines the Traits UI view for a PlotAxis """
-
-import six
-
 from traits.api import TraitError
 from traitsui.api import View, HGroup, Group, VGroup, Item, TextEditor
 
@@ -14,7 +11,7 @@ def float_or_auto(val):
     try:
         return float(val)
     except:
-        if isinstance(val, six.string_types) and val == "auto":
+        if isinstance(val, str) and val == "auto":
             return val
     raise TraitError("Tick interval must be a number or 'auto'.")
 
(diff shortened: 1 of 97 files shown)


chaco/axis_view.py
```diff
@@ -32,11 +29,8 @@ def float_or_auto(val):
                     label="Main"),
                 Group(
                     Item("tick_color", label="Color", style="custom"),
-                         #editor=EnableRGBAColorEditor()),
```
rahulporuri, line 35: i'm happy when dead code is removed - but it'd be nice to mention this in the PR description.

rahulporuri (commented):
I want to take a second look at the changes from `sm.zip` to `zip`. Some of them feel off and I don't trust the test suite to catch all potential issues.

aaronayres35:
> I want to take a second look at the changes from `sm.zip` to `zip`. Some of them feel off and I don't trust the test suite to catch all potential issues.

I was following here: https://six.readthedocs.io/
Specifically in the supported renames section

chaco/legend.py
```diff
@@ -358,7 +355,7 @@ def get_preferred_size(self):
         if len(self.plots) == 0:
             return [0, 0]
 
-        plot_names, visible_plots = list(sm.map(list, sm.zip(*sorted(self.plots.items()))))
+        plot_names, visible_plots = list(map(list, zip(*sorted(self.plots.items()))))
```
rahulporuri, line 358: I'm not a big fan of such code - `list(map(list, zip(*sorted(self.plots.items()))))` - because i find it hard to understand. Anyway, that's not an actionable item in this PR given the lack of context.

rahulporuri (approved):
LGTM

chaco/legend.py
aaronayres35: I agree, any code with "`)))))`" smells a bit, but that is a problem for another day/PR

PR #555, opened 2021-01-06
Remove python version checks

aaronayres35:
part of #483 

This PR removes a few checks for a specific python version and only keeps code specific to python 3+.

Based on this comment: https://github.com/enthought/chaco/blob/4a6683521cce012b5df4830d64f41518798fc817/docs/scipy_tutorial/Scripts/shBrushPython.js#L1
I suspect we Can remove this .js file entirely?

### DIFF
--- a/chaco/contour/cntr.c
+++ b/chaco/contour/cntr.c
@@ -1700,19 +1700,14 @@ static PyMethodDef module_methods[] = {
 };
 
 
-#if PY_MAJOR_VERSION >= 3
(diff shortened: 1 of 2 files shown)


kitchoi:
> Based on this comment:
>
> chaco/docs/scipy_tutorial/Scripts/shBrushPython.js
 > /* Python 2.3 syntax contributed by Gheorghe Milas */ 
>
> I suspect we Can remove this .js file entirely?

Just the comment alone is not compelling enough for me to think removing it is safe. So I'd prefer not.

kitchoi (approved):

PR #565, opened 2021-02-25
Remove redundant 'control' in ZoomTool

notmatthancock:
The 'control' option is repeated, so I removed the second one.


rahulporuri (approved):
Thanks @notmatthancock ! LGTM

PR #566, opened 2021-03-09
Add ci command for building documentation

aaronayres35:
this PR simply pulls the `docs` ci command from other ets projects into chaco.

### DIFF
--- a/ci/edmtool.py
+++ b/ci/edmtool.py
@@ -108,6 +108,15 @@ extra_dependencies = {
     'null': set()
 }
 
+doc_dependencies = {
+    "sphinx",
+    "enthought_sphinx_theme"
+}
+
+doc_ignore = {
+    "*/tests",
+}
+
 environment_vars = {
     'pyside2': {'ETS_TOOLKIT': 'qt4', 'QT_API': 'pyside2'},
     'pyqt': {'ETS_TOOLKIT': 'qt4', 'QT_API': 'pyqt'},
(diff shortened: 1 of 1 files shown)


rahulporuri (approved):
LGTM. This is a good place to start. There a few issues, two of which needs to be addressed in this PR

- it looks like the api docs are autogenerated but they aren't ignored in `.gitignore` so git sees a lot of changed/new files. <- This is the only thing that needs to be fixed in this PR.
- we're building api docs but we're not linking to them in the built docs <- this too needs to be addressed in this PR. IMO.
- we are installing `enthought_sphinx_theme` but we aren't using it for the docs
- the version information seems to be hardcoded in the sphinx conf which is why i see docs for chaco 4.8

ci/edmtool.py
```diff
@@ -267,6 +276,61 @@ def update(runtime, toolkit, environment):
...
+    execute(commands, parameters)
+    click.echo("Done installing documentation tools")
+
+    click.echo(
+        "Regenerating API docs in  '{environment}'".format(**parameters)
+    )
+    api_path = os.path.join("docs", "source", "api")
+    if os.path.exists(api_path):
+        rmtree(api_path)
+    os.makedirs(api_path)
```
rahulporuri, line 307: I think this is borrowed from traitsui or pyface. Can you borrow the code from traits or traits-futures instead? It will be nice if we can borrow the sphinx api templates.

PR #462, opened 2019-07-13
Documentation Refactor

corranwebster:

PR #567, opened 2021-03-09
Update various comments and docstrings

aaronayres35:
part of an effort to break #462 down into various smaller PRs

This PR simply pulls changes directly from that PR.  Namely, those changes involving simply updating example file docstrings, or updating comments on trait definitions


rahulporuri (approved):
LGTM

PR #572, opened 2021-03-10
Update api docs

aaronayres35:
part of an effort to break #462 down into various smaller PRs

This PR cleans up and adds to the existing api documentation.  It also adds an `index.rst` file for the api docs that is currently unused here.  It will be used in 462, but I figured is fine to go in alone for now.  

The api docs are currently manually constructed / organized, which may very well be what we want to do.  We could instead auto build the api docs entirely, however then the organization will mirror that of the package itself.  This may be more confusing for documentation purposes.  However, if it is more confusing, that may suggest that the modules in the codebase could use some reorganization?  In any case, this is not relevant to this PR, and can be addressed later.

### DIFF
--- a/docs/source/api/axis.rst
+++ b/docs/source/api/axis.rst
@@ -3,7 +3,7 @@
 Axis and Grid
 --------------
 
-.. currentmodule:: chaco.api
+.. currentmodule:: chaco.axis
 
 :class:`PlotAxis`
 ==========================
@@ -11,17 +11,28 @@ Axis and Grid
     :members:
     :show-inheritance:
 
+:class:`MinorPlotAxis`
+==========================
+.. autoclass:: MinorPlotAxis
+    :members:
+    :show-inheritance:
+.. currentmodule:: chaco.label_axis
+
 :class:`LabelAxis`
 ==========================
 .. autoclass:: LabelAxis
     :members:
     :show-inheritance:
 
+.. currentmodule:: chaco.grid
+
 :class:`PlotGrid`
 ==========================
 .. autoclass:: PlotGrid
     :members:
     :show-inheritance:
+
+.. currentmodule:: chaco.ticks
 
 :class:`AbstractTickGenerator`
 ===============================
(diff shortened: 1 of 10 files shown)


rahulporuri:
The thing i find useful about the current manually curated code is the fact that it brings together separate modules which implement the same interface e.g. all data ranges, plot factories, renderers etc - which is actually incredibly useful knowledge.

Given that, I think this PR needs to go in - but we also need to figure out a better way to mix autogenerated API docs with such curated collections.

rahulporuri:
> The thing i find useful about the current manually curated code is the fact that it brings together separate modules which implement the same interface e.g. all data ranges, plot factories, renderers etc - which is actually incredibly useful knowledge.

Having said that but looked at the PR a bit longer, I think it'll be better if we create normal user guide docs e.g. `docs/source/user_manual/axis` instead of `docs/source/api/axis` - where we could link the relevant classes. The documentation page will also be less verbose/cluttered with information than it is at the moment because it'll be a simple list of links to the relevant objects.

aaronayres35:
> > The thing i find useful about the current manually curated code is the fact that it brings together separate modules which implement the same interface e.g. all data ranges, plot factories, renderers etc - which is actually incredibly useful knowledge.
> 
> Having said that but looked at the PR a bit longer, I think it'll be better if we create normal user guide docs e.g. `docs/source/user_manual/axis` instead of `docs/source/api/axis` - where we could link the relevant classes. The documentation page will also be less verbose/cluttered with information than it is at the moment because it'll be a simple list of links to the relevant objects.

That makes sense.  We also have `docs/source/user_manual/modules_and_classes.rst` in #462 that has a similar theme.  I'm not sure what the original idea behind that doc and its relationship / difference from the api docs was intended to be.

aaronayres35:
I think I am leading towards pushing this PR through for now. The changes in this PR are pretty minor / clean up.  They aren't really related to the overall plan of what we are going to do about api docs.  Bigger picture we will want to refactor how we handle curated api docs, generated api docs, and the user manual itself.  Nonetheless, standalone I think this PR makes an incremental improvement to what exists currently, and we can merge.  Even if long term we decide to remove the relevant files altogether, or whatever we end up doing

@rahulporuri thoughts?

PR #574, opened 2021-03-18
remove SelectableOverlayPlotContainer and references to it

### LINKED ISSUE #573: SelectableOverlayPlotContainer can be removed
This is an _old_ class that has been superceded by the use of overlays.

It does not appear to be being used and can be safely removed for 5.0.

aaronayres35:
closes #573 

This PR simply remove the `SelectableOverlayPlotContainer` class and references to it (api module and docs)

Note that this was exposed in the api, but per the issue we believe it can be safely removed


rahulporuri (approved):
LGTM. Again, I don't see any meaningful users [on GitHub](https://github.com/search?p=1&q=selectable_overlay_container&type=Code)

PR #575, opened 2021-03-18
remove Serializable mixin and related code

### LINKED ISSUE #250: Serializable mixin should be removed
The class is not used by any current code, appears to be broken, and the problem it was meant to solve (selection of which traits to pickle) is better solved via the use of `transient` traits metadata.

Technically this is a backwards-incompatible change, however.

aaronayres35:
closes #250 

This PR removes the `Serializable` mixin class and related tests.  It also removes any definitions of a `_post_load` method in various classes as the only place I saw it originally defined / used was `Serializable` (including a search of enable and traits).


chaco/grid.py
```diff
@@ -407,8 +407,6 @@ def _orientation_changed(self):
 
 
     ### Persistence ###########################################################
-    #_pickles = ("orientation", "line_color", "line_style", "line_weight",
-    #            "grid_interval", "mapper")
 
     def __getstate__(self):
```
rahulporuri, line 411: I don't know if the `__getstate__` methods can also be thrown out if the serialization code is being thrown out.

aaronayres35:
closing in favor of #598

PR #576, opened 2021-03-18
remove chaco.plot_canvas and related code

### LINKED ISSUE #254: Remove chaco.plot_canvas and related code
This code is not currently used and dates from Chaco 1, I think.  There is an example of its use in the demo code, but it appears old and broken (at least on OS X).

This should probably be removed, although such a change is technically not backwards compatible.

aaronayres35:
closes #254 

This PR removes `chaco.plot_canvas`, the related `chaco.plot_canvas_toolbar` and the canvas example from the code base


rahulporuri (dismissed):
LGTM. Like with #577 , I dont think this will be a disruptive change but it does look like there are some users of `chaco.plot_canvas` - for them, it'll be useful to know what they can move to and how.

- [github search for `chaco.plot_canvas`](https://github.com/search?q=chaco.plot_canvas&type=Code)
- [github search for `chaco.plot_canvas_toolbar`](https://github.com/search?p=1&q=chaco.plot_canvas_toolbar&type=Code)

corranwebster:
As discussed in the original issue - `PlotCanvas` may not be a problem, but the demo of it _is_.

PR #578, opened 2021-03-18
have ArrayDataSource.get_data return empty(shape=0,)), not 0.0 if _data is None

### LINKED ISSUE #255: ArrayDataSource get_data() returns scalar 0.0 when _data is None
The `ArrayDataSource` is meant to represent a 1D data structure stored in an array.  For some reason, when no data value has been set, the `get_data()` method returns a scalar `0.0` value.

It would probably make most sense for the value to be `empty(shape=(0,))` (ie. a 1D array of length 0).

Changing this may break existing code which expects empty `ArrayDataSources` to return a scalar 0.  On the flip side, this will potentially simplify the logic for `get_bounds()` and similar.

aaronayres35:
closes #255 

This PR makes `ArrayDataSource.get_data` return `empty(shape=0,))` not 0.0 if `_data` is `None`.

Note this is a potentially breaking change if code depended on getting 0.0 by default


chaco/array_data_source.py
```diff
@@ -248,11 +248,8 @@ def _compute_bounds(self, data=None):
             #data = self._data
```
rahulporuri, line 248: looks like there's some unnecessary commented out code we can remove here.

rahulporuri (approved):
LGTM

rahulporuri:
@aaronayres35 looks like you'll have to merge master in after you merge the github actions PR.

PR #582, opened 2021-03-24
Remove commented out code

aaronayres35:
This PR simply removes a few sections of commented out code

### DIFF
--- a/chaco/data_range_1d.py
+++ b/chaco/data_range_1d.py
@@ -98,9 +98,6 @@ class DataRange1D(BaseDataRange):
     # The actual numerical value for the high setting.
     _high_value = CFloat(inf)
 
-    # A list of attributes to persist
-    # _pickle_attribs = ("_low_setting", "_high_setting")
-
     #------------------------------------------------------------------------
     # AbstractRange interface
     #------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


rahulporuri (approved):
LGTM. Not sure what `_pickles` is.

rahulporuri:
Note that there are a couple of other instances of `_pickle*` that are commented which can also be removed. And if i understand correctly, we can also remove the persistence related methods - `__getstate__` and `__setstate__`.

I tried digging and at the bottom of the search - i still find these lines commented. I don't know which PR/why these lines were commented out.

corranwebster:
> LGTM. Not sure what `_pickles` is.

It looks like it's used by the `Serializable` class: https://github.com/enthought/chaco/blob/e3423614d654a203bd82ee087c085abe13b465b1/chaco/serializable.py which even has tests https://github.com/enthought/chaco/blob/e3423614d654a203bd82ee087c085abe13b465b1/chaco/tests/serializable_base.py but isn't used anywhere in Chaco.

Modern usage would be to mark traits as `transient` if you _don't_ want them serialized (which is automatic for property and delegate traits).

The whole system can probably go.

corranwebster:
I've added #583 for dealing with `__getstate__` and `__setstate__`.

The `_post_load` methods are more interesting, because they hint at things which might go into a `traits_init` function and allow simplification of `__init__` functions, but that's a bigger and more complex job.

PR #585, opened 2021-03-30
Replace on_trait_change decorators with observe

aaronayres35:
This PR replaces uses of the `@on_trait_change` decorator with the equivalent form using `observe`.

I made changes one file at a time and ran test suite / tested examples after each change.  However, it is possible certain changed code is not under test.

### DIFF
--- a/chaco/barplot.py
+++ b/chaco/barplot.py
@@ -4,7 +4,7 @@ import logging
 
 from numpy import array, compress, column_stack, invert, isnan, transpose, zeros
 from traits.api import Any, Bool, Enum, Float, Instance, Property, \
-        Range, Tuple, cached_property, observe
+        Range, Tuple, cached_property
 from enable.api import black_color_trait
 from kiva.constants import FILL_STROKE
 
@@ -69,11 +69,11 @@ class BarPlot(AbstractPlotRenderer):
     antialias = Bool(True)
 
     #: Width of the border of the bars.
-    line_width = Float(1.0)
+    line_width = Float(1.0, requires_redraw=True)
     #: Color of the border of the bars.
-    line_color = black_color_trait
+    line_color = black_color_trait(requires_redraw=True)
     #: Color to fill the bars.
-    fill_color = black_color_trait
+    fill_color = black_color_trait(requires_redraw=True)
 
     #: The RGBA tuple for rendering lines. It is always a tuple of length 4.
     #: It has the same RGB values as :attr:`line_color`, and its alpha value
@@ -86,7 +86,7 @@ class BarPlot(AbstractPlotRenderer):
     effective_fill_color = Property(Tuple, depends_on=['fill_color', 'alpha'])
 
     #: Overall alpha value of the image. Ranges from 0.0 for transparent to 1.0
-    alpha = Range(0.0, 1.0, 1.0)
+    alpha = Range(0.0, 1.0, 1.0, requires_redraw=True)
 
 
     #use_draw_order = False
@@ -414,11 +414,6 @@ class BarPlot(AbstractPlotRenderer):
 
         self.invalidate_draw()
         self._cache_valid = False
-
-    @observe('line_color, line_width, fill_color, alpha')
-    def _attributes_changed(self, event):
-        self.invalidate_draw()
-        self.request_redraw()
 
     def _bounds_changed(self, old, new):
         super(BarPlot, self)._bounds_changed(old, new)
--- a/chaco/barplot.py
+++ b/chaco/barplot.py
@@ -4,7 +4,7 @@ import logging
 
 from numpy import array, compress, column_stack, invert, isnan, transpose, zeros
 from traits.api import Any, Bool, Enum, Float, Instance, Property, \
-        Range, Tuple, cached_property, on_trait_change
+        Range, Tuple, cached_property, observe
 from enable.api import black_color_trait
 from kiva.constants import FILL_STROKE
 
@@ -415,8 +415,8 @@ class BarPlot(AbstractPlotRenderer):
         self.invalidate_draw()
         self._cache_valid = False
 
-    @on_trait_change('line_color, line_width, fill_color, alpha')
-    def _attributes_changed(self):
+    @observe('line_color, line_width, fill_color, alpha')
+    def _attributes_changed(self, event):
         self.invalidate_draw()
         self.request_redraw()
 
--- a/chaco/base_xy_plot.py
+++ b/chaco/base_xy_plot.py
@@ -75,7 +75,7 @@ class BaseXYPlot(AbstractPlotRenderer):
     orientation = Enum("h", "v")
 
     #: Overall alpha value of the image. Ranges from 0.0 for transparent to 1.0
-    alpha = Range(0.0, 1.0, 1.0)
+    alpha = Range(0.0, 1.0, 1.0, requires_redraw=True)
 
     #------------------------------------------------------------------------
     # Convenience readonly properties for common annotations
--- a/chaco/color_bar.py
+++ b/chaco/color_bar.py
@@ -4,8 +4,10 @@
 from numpy import array, arange, ascontiguousarray, ones, transpose, uint8
 
 # Enthought library imports
-from traits.api import Any, Bool, Enum, Instance, Property, \
-                                 cached_property, on_trait_change
+from traits.api import (
+    Any, Bool, Enum, Instance, Property, cached_property, observe
+)
+from traits.observation.api import parse
 from kiva.image import GraphicsContext
 
 # Local imports
@@ -217,8 +219,8 @@ class ColorBar(AbstractPlotRenderer):
     def _updated_changed_for_color_mapper(self):
         self._update_mappers()
 
-    @on_trait_change('[index_mapper,color_mapper].+')
-    def _either_mapper_changed(self):
+    @observe(parse('[index_mapper,color_mapper]').match(lambda n, t: True))
+    def _either_mapper_changed(self, event=None):
         self.invalidate_draw()
         self.request_redraw()
 
--- a/chaco/color_bar.py
+++ b/chaco/color_bar.py
@@ -220,7 +220,7 @@ class ColorBar(AbstractPlotRenderer):
         self._update_mappers()
 
     @observe(parse('[index_mapper,color_mapper]').match(lambda n, t: True))
-    def _either_mapper_changed(self, event=None):
+    def _either_mapper_updated(self, event=None):
         self.invalidate_draw()
         self.request_redraw()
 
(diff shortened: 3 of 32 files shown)


chaco/barplot.py
```diff
@@ -415,8 +415,8 @@ def _update_mappers(self):
         self.invalidate_draw()
         self._cache_valid = False
 
-    @on_trait_change('line_color, line_width, fill_color, alpha')
-    def _attributes_changed(self):
+    @observe('line_color, line_width, fill_color, alpha')
+    def _attributes_changed(self, event):
```
rahulporuri, line 419: i'm not a fan of the `*_changed` method name especially when it's not a trait change handler but i can't think of one better either.

rahulporuri: In general, i've been told that the methods shouldn't be named based on what changed - rather, they should be named to convey what they are going to do.

Until we find a better name, maybe we can just call this `_attributes_updated` - which i've seen elsewhere in this PR.

chaco/color_bar.py
```diff
@@ -217,8 +219,8 @@ def _updated_changed_for_index_mapper(self):
     def _updated_changed_for_color_mapper(self):
         self._update_mappers()
 
-    @on_trait_change('[index_mapper,color_mapper].+')
-    def _either_mapper_changed(self):
+    @observe(parse('[index_mapper,color_mapper]').match(lambda n, t: True))
+    def _either_mapper_changed(self, event=None):
```
rahulporuri, line 223: let's update this too to `_either_mapper_updated` or `_invalidate_and_redraw`

chaco/barplot.py
rahulporuri: alternatively, we could rename it to `_invalidate_and_redraw` - if no such method already exists in the class heirarchy - because that's what the method does.

chaco/colormapped_scatterplot.py
```diff
@@ -413,8 +411,8 @@ def _color_mapper_changed(self, old, new):
 
         return
 
-    @on_trait_change('color_mapper:updated')
-    def _color_mapper_updated(self):
+    @observe('color_mapper:updated')
+    def _color_mapper_updated(self, event):
```
rahulporuri, line 415: same comment as above about updating method name.

chaco/data_range_2d.py
```diff
@@ -227,6 +227,6 @@ def _sources_changed(self, old, new):
         self._yrange.sources = [s._ydata for s in self.sources]
         self.refresh()
 
-    @on_trait_change("_xrange.updated,_yrange.updated")
-    def _subranges_updated(self):
+    @observe("_xrange.updated,_yrange.updated")
+    def _subranges_updated(self, event):
```
rahulporuri, line 231: this could just be `_set_updated_trait` or `_set_updated`

chaco/polygon_plot.py
```diff
@@ -144,8 +143,8 @@ def hittest(self, screen_pt, threshold=7.0, return_distance=False):
     # Event handlers
     #------------------------------------------------------------------------
 
-    @on_trait_change('edge_color, edge_width, edge_style, face_color, alpha')
-    def _attributes_changed(self):
+    @observe('edge_color, edge_width, edge_style, face_color, alpha')
+    def _attributes_changed(self, event):
```
rahulporuri, line 147: same comment as above - about renaming this to `_invalidate_and_redraw` if such a method doesn't already exist in the class heirarchy

chaco/segment_plot.py
```diff
@@ -351,11 +350,11 @@ def _render_icon(self, gc, x, y, width, height):
             gc.move_to(x, y)
             gc.line_to(width, height)
 
-    @on_trait_change(
+    @observe(
         'alpha, color_data:data_changed, color_mapper:updated, '
         'width_data:data_changed, width_mapper.updated, +redraw'
```
rahulporuri, line 355: does observe understand `+redraw`?

chaco/segment_plot.py
```diff
@@ -351,11 +350,11 @@ def _render_icon(self, gc, x, y, width, height):
...
     )
-    def _attributes_changed(self):
+    def _attributes_changed(self, event):
```
rahulporuri, line 357: same comment as above about renaming this to `_invalidate_and_redraw` if possible

chaco/toolbar_plot.py
```diff
@@ -51,12 +50,13 @@ def _bounds_changed(self, old, new):
         self.toolbar.do_layout(force=True)
         super(ToolbarPlot, self)._bounds_changed(old, new)
 
-    @on_trait_change('toolbar')
-    def _toolbar_changed(self, name, obj, old, new):
+    @observe('toolbar')
+    def _on_toolbar_changed(self, event):
```
rahulporuri, line 54: Maybe rename this to `_update_toolbar` if possible?

Also, it might be cleaner if we can create aliases `old, new = event.old, event.new` at the top of this method.

examples/demo/advanced/data_cube.py
```diff
@@ -66,7 +67,15 @@ def __init__(self, *args, **kwargs):
...
+    @observe(
+        match(
+            lambda name, trait: (
+                name.startswith("npts_") or
+                name.startswith("min_") or
+                name.startswith("max_")
+                )
+            )
+    )
     def compute_model(self):
```
rahulporuri, line 79:
```suggestion
    def compute_model(self, event=None):
```

examples/demo/advanced/data_cube.py
```diff
@@ -222,7 +231,7 @@ def _wheel_callback(self, tool, wheelamt):
         self.bottom.invalidate_and_redraw()
         return
 
-    def _create_window(self):
+    def _create_component(self):
```
rahulporuri, line 234: this looks like an unrelated change

examples/demo/advanced/data_cube.py
```diff
@@ -272,7 +281,7 @@ def _create_window(self):
         container.add(bottomplot)
 
         self.container = container
-        return Window(self, -1, component=container)
+        return container
```
rahulporuri, line 284: same here - this too looks like an unrelated change

chaco/tools/toolbars/plot_toolbar.py
```diff
@@ -282,12 +281,12 @@ def _calculate_height(self):
...
-    def _hiding_changed(self):
+    @observe('hiding')
+    def _on_hiding_changed(self, event):
         self._layout_needed = True
         self.request_redraw()
 
-    @on_trait_change('auto_hide')
-    def _auto_hide_changed(self):
+    @observe('auto_hide')
+    def _on_auto_hide_changed(self, event):
```
rahulporuri, line 290: I wish we could come up with better names for `_on_hiding_changed` and `_on_auto_hide_changed`.

Maybe `_set_layout_needed_and_redraw` and `_set_hiding_and_redraw`?

The names sound kinda silly now because the code in the methods almost matches up with the method names.

rahulporuri (commented):
almost LGTM

aaronayres35:
Okay I have gone through and done a bunch of renaming and also started to follow Corran's suggestion about using metadata and pulling the hook for `invalidate_and_redraw` up higher.  I have pulled in into `PlotComponent` for now so that it could all be done in this PR.  Eventually we may want to pull that even further up into `Component` in enable.

Now if a particular trait changing will require an invalidate and redraw, one can simply set the metadata `requires_redraw=True` in the trait definition.
Note I did not do this in _all_ such cases here, because sometimes we are setting these things up on nested traits. eg all traits on a color mapper.  But I imagine these objects could be used in other contexts and we dont really want to mess with trait definitions that don't belong to that particular object that wants invalidate_and_redraw called?
I am still a little unsure about this and would like to discuss a bit more.

Nonetheless, I believe this PR is close to being ready to push through

chaco/base_xy_plot.py
```diff
@@ -75,7 +75,7 @@ class BaseXYPlot(AbstractPlotRenderer):
     orientation = Enum("h", "v")
 
     #: Overall alpha value of the image. Ranges from 0.0 for transparent to 1.0
-    alpha = Range(0.0, 1.0, 1.0)
+    alpha = Range(0.0, 1.0, 1.0, requires_redraw=True)
```
rahulporuri, line 78: This will now spill over into other classes which inherit from `BaseXYPlot` i.e. any subclass of `BaseXYPlot` will invalidate and redraw if the alpha value changes.. That might be the right behavior but we need to be conscious of the fact that we might be modifying behavior.

rahulporuri (approved):
Mostly LGTM with a few comments - 

1. I would have preferred it if we made the `request_redraw` related changes in a separate PR - given that those changes are adding a new feature and are sort of unrelated to the scope of this PR.
2. We need to document the fact that classes which inherit from `PlotComponent` can define traits with the metadata `request_redraw` if they want changes to the trait to trigger an invalidate and redraw.
3. I'm trying to think of a different/better name than `request_redraw` but i can't come up with any at the moment. We want to be conscious of the choice because it will be part of the public API.

aaronayres35:
> Mostly LGTM with a few comments -
> 
> 1. I would have preferred it if we made the `request_redraw` related changes in a separate PR - given that those changes are adding a new feature and are sort of unrelated to the scope of this PR.

Yeah you are right, my bad on that.  I can still pull it out at this point if you prefer 

> 2. We need to document the fact that classes which inherit from `PlotComponent` can define traits with the metadata `request_redraw` if they want changes to the trait to trigger an invalidate and redraw.

That makes sense, I will add this

> 3. I'm trying to think of a different/better name than `request_redraw` but i can't come up with any at the moment. We want to be conscious of the choice because it will be part of the public API.

I believe it is currently `requires_redraw` not `request_redraw` (perhaps request is better though 🤔 ).  A longer winded but more specific alternative might be `requires_redraw_on_change`? I am not sure, naming things is hard.

rahulporuri:
> Yeah you are right, my bad on that. I can still pull it out at this point if you prefer

No no lets not.

> I believe it is currently requires_redraw not request_redraw (perhaps request is better though 🤔 ). A longer winded but more specific alternative might be requires_redraw_on_change? I am not sure, naming things is hard.

Let's go with `requires_redraw` for now. we can revisit this before the release and put it up for vote with the rest of the team/folks.

aaronayres35:
Orthogonal to this PR, but we should add a "maintenance" news fragment type.  I wasn't sure which of "feature, bugfix, deprecation, removal, doc, test, build" to classify this PR as, but it feels like it will be ultimately relevant for the changelog (as I imagine 5.1 of apptools main purpose will be this migration to observe / depending on traits 6.2)

I am going to call it a feature for now simply because IDK where to put it.  When generating the actual changelog we will want to move this, but this way it will still be picked up

PR #589, opened 2021-03-30
return correct sized tuples of all 0s not None if image plot resized too small

### LINKED ISSUE #528: ImagePlot.draw generates a fatal error when the window showing a plot was resized too small
**Problem Description**
I have a plot shown in a window. There are a few tools added to the plot: a hovering tool, a selection tool, a zoom tool. When I resize the window so small that the plot does not show the application crashes. According to the traceback the error comes from Chaco.

**Reproduction Steps:**

I tried to reproduce with simple plots, but I cannot. My first thought is that the bug is related to the presence of the tools. I can share the traceback

```
  File "lib/python3.6/site-packages/enable/qt4/base_window.py", line 213, in paintEvent
    self.handler.paintEvent(event)
  File "lib/python3.6/site-packages/enable/qt4/base_window.py", line 65, in paintEvent
    self._enable_window._paint(event)
  File lib/python3.6/site-packages/enable/abstract_window.py", line 471, in _paint
    self.component.draw(gc, view_bounds=(0, 0, size[0], size[1]))
  File "lib/python3.6/site-packages/enable/component.py", line 427, in draw
    self._draw(gc, view_bounds, mode)
  File "lib/python3.6/site-packages/enable/component.py", line 779, in _draw
    self._dispatch_draw(layer, gc, view_bounds, mode)
  File "lib/python3.6/site-packages/enable/container.py", line 273, in _dispatch_draw
    component._dispatch_draw(layer, gc, new_bounds, mode)
  File "lib/python3.6/site-packages/enable/component.py", line 799, in _dispatch_draw
    handler(gc, view_bounds, mode)
  File "lib/python3.6/site-packages/chaco/base_2d_plot.py", line 204, in _draw_image
    self._render(gc)
  File "lib/python3.6/site-packages/chaco/image_plot.py", line 124, in _render
    self._compute_cached_image()
  File "lib/python3.6/site-packages/chaco/image_plot.py", line 255, in _compute_cached_image
    col_min, col_max, row_min, row_max = index_bounds
TypeError: 'NoneType' object is not iterable
Fatal Python error: Aborted
```

**Expected behavior:**

I would expect the plot to simply not show when the window is too small and that it would not crash my application. I would also expect that the tools won't be available.

The bug can be reproduced on Linux and MacOS, Python 3.6

aaronayres35:
fixes #528 (I think)

With the change made in this PR I no longer see the crash described here: https://github.com/enthought/chaco/issues/528#issuecomment-809297495

This PR simply updates `_calc_zoom_coords` to return what it claims to return:
```
Returns
-------
    index_bounds : 4-tuple
        The column and row indices (col_min, col_max, row_min, row_max) of
        the sub-image to be extracted and drawn into `screen_rect`.
    screen_rect : 4-tuple
        (x, y, width, height) rectangle describing the pixels bounds where
        the image will be rendered in the plot.
```

In this particular case if we have zoomed in so far that the image has either length or width 0, the the screen rect showing it is effectively non-existent, and there is no data to show in it.  
Technically `((0, 0, 0, 0), (0, 0, 0, 0))` isn't really the exactly correct thing to return here (instead you could think that screen_rect should be a 0x0 rectangle at the point where the image should be on screen, not at x=0, y=0).  However, there is actually nothing to draw so it effectively doesn't matter.

The only place this method is called is by ` _compute_cached_image`.  If `_compute_cached_image` gets ``((0, 0, 0, 0), (0, 0, 0, 0))`` everything works out as expected as we end up with `data=data[0:0, 0:0]` making `data.shape = (0, 0, 3)` and a screen_rect that is 0x0.

Note: now when I slowly shrink the window once I hit the critical point rather than the example crashing, I see:
```
/Users/aayres/Desktop/chaco/chaco/axis.py:613: RuntimeWarning: invalid value encountered in true_divide
  self._axis_pixel_vector = self._axis_vector/sqrt(dot(self._axis_vector,self._axis_vector))
  ```
This is not new from the changes in this PR (was masked by the crash previously), but occurs because we end up with an `_axis_vector` of (0, 0) when we shrink that far

### DIFF
--- a/chaco/tests/test_image_plot.py
+++ b/chaco/tests/test_image_plot.py
@@ -2,7 +2,6 @@ import os
 
 import tempfile
 from contextlib import contextmanager
-from scipy.misc import face
 
 import unittest
 
@@ -181,7 +180,7 @@ class TestResultImage(unittest.TestCase):
             )
 
             def _plot_default(self):
-                pd = ArrayPlotData(image=face())
+                pd = ArrayPlotData(image=RGB)
                 plot = Plot(pd)
                 plot.img_plot('image')
 
--- a/chaco/tests/test_image_plot.py
+++ b/chaco/tests/test_image_plot.py
@@ -166,6 +166,7 @@ class TestResultImage(unittest.TestCase):
                                  origin='bottom right', orientation='v')
 
     # regression test for enthought/chaco#528
+    @unittest.skipIf(ETSConfig.toolkit=='null', "Skip on 'null' toolkit")
     def test_resize_to_zero(self):
         class TestResize(HasTraits):
             plot = Instance(Component)
--- a/chaco/tests/test_image_plot.py
+++ b/chaco/tests/test_image_plot.py
@@ -2,14 +2,27 @@ import os
 
 import tempfile
 from contextlib import contextmanager
+from scipy.misc import face
 
 import unittest
 
 import numpy as np
 
+from enable.api import Component, ComponentEditor
+from traits.api import HasTraits, Instance
 from traits.etsconfig.api import ETSConfig
-from chaco.api import (PlotGraphicsContext, GridDataSource, GridMapper,
-                       DataRange2D, ImageData, ImagePlot)
+from traitsui.api import Group, UItem, View
+from traitsui.testing.api import UITester
+from chaco.api import (
+    ArrayPlotData,
+    PlotGraphicsContext,
+    GridDataSource,
+    GridMapper,
+    DataRange2D,
+    ImageData,
+    ImagePlot,
+    Plot
+)
 
 
 # The Quartz backend rescales pixel values, so use a higher threshold.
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by rahulporuri** [comment]
I tested this locally by checking the example https://github.com/enthought/chaco/blob/174342ff70f78941f011656cc994a9c28463c547/examples/demo/image_plot_origin_and_orientation.py - and that example doesn't crash on this branch - whereas the example crashes on the main branch.

But, I don't like the fact that I can squeeze the window horizontally and vertically into an absurd state. But, an working application in an absurd state is better than a crashed application i guess?

why-chaco-why
tall

**Comment by rahulporuri** [comment]
> This is not new from the changes in this PR (was masked by the crash previously), but occurs because we end up with an _axis_vector of (0, 0) when we shrink that far

I'm not sure why an `_axis_vector` would raise that runtime warning. Did you mean `_axis_pixel_vector`? It'd be useful to investigate what is actually causing that runtime warning - and maybe open an issue for it.

**Review by rahulporuri — APPROVED** [review]
Code changes LGTM. It'd be awesome if you could put together a regression test for this change, that would be awesome.

Let's wait for the devs who reported the issue to also get back to us.

**Comment by rahulporuri** [comment]
@jvkersch any chance you can take a few minutes to look at this PR and the issue #528 ?

**Comment by sallenEnth** [comment]
@aaronayres35 @rahulporuri I tested and I don't see the bug anymore. Thank you for the fix.

File: chaco/tests/test_image_plot.py [hunk 1]
```diff
@@ -152,3 +165,31 @@ def test_vertical_bottom_right(self):
...
+                plot.img_plot('image')
+
+                return plot
+        
+        test_obj = TestResize()
+        tester = UITester()
+
+         # should not fail
+        with tester.create_ui(test_obj) as ui:
+            pass
```
aaronayres35, line 195: this test is a little ugly namely from the 101 magic number.  I was unsure how to better write this test.

Without the fix this test fails, with it it passes.

rahulporuri (approved):
Still LGTM

PR #590, opened 2021-03-31
Replace dynamic on_trait_change with observe

aaronayres35:
This PR starts replacing the dynamic uses of on_trait_change with observe equivalents

I am going to arbitrarily stop here to keep this PR shorter to review.  After these changes, there are 18 files left containing `.on_trait_change` which can be addressed in a separate PR.

### DIFF
--- a/chaco/base_2d_plot.py
+++ b/chaco/base_2d_plot.py
@@ -325,10 +325,10 @@ class Base2DPlot(AbstractPlotRenderer):
 
     def _value_changed(self, old, new):
         if old is not None:
-            old.on_trait_change(self._update_value_data,
+            old.observe(self._update_value_data,
                                 "data_changed", remove=True)
         if new is not None:
-            new.on_trait_change(self._update_value_data, "data_changed")
+            new.observe(self._update_value_data, "data_changed")
         self._update_value_data()
 
     def _index_mapper_changed(self, old, new):
--- a/chaco/base_2d_plot.py
+++ b/chaco/base_2d_plot.py
@@ -89,14 +89,11 @@ class Base2DPlot(AbstractPlotRenderer):
         self.trait_set(**kwargs_tmp)
         super(Base2DPlot, self).__init__(**kwargs)
         if self.index is not None:
-            self.index.observe(self._update_index_data,
-                                       "data_changed")
+            self.index.observe(self._update_index_data, "data_changed")
         if self.index_mapper:
-            self.index_mapper.observe(self._update_index_mapper,
-                                              "updated")
+            self.index_mapper.observe(self._update_index_mapper, "updated")
         if self.value is not None:
-            self.value.observe(self._update_value_data,
-                                       "data_changed")
+            self.value.observe(self._update_value_data, "data_changed")
         # If we are not resizable, we will not get a bounds update upon layout,
         # so we have to manually update our mappers
         if self.resizable == "":
@@ -317,25 +314,21 @@ class Base2DPlot(AbstractPlotRenderer):
 
     def _index_changed(self, old, new):
         if old is not None:
-            old.obseve(self._update_index_data,
-                                "data_changed", remove=True)
+            old.obseve(self._update_index_data, "data_changed", remove=True)
         if new is not None:
             new.obseve(self._update_index_data, "data_changed")
         self._update_index_data()
 
     def _value_changed(self, old, new):
         if old is not None:
-            old.observe(self._update_value_data,
-                                "data_changed", remove=True)
+            old.observe(self._update_value_data, "data_changed", remove=True)
         if new is not None:
             new.observe(self._update_value_data, "data_changed")
         self._update_value_data()
 
     def _index_mapper_changed(self, old, new):
         if old is not None:
-            old.observe(self._update_index_mapper,
-                                "updated", remove=True)
+            old.observe(self._update_index_mapper, "updated", remove=True)
         if new is not None:
             new.observe(self._update_index_mapper, "updated")
         self._update_index_mapper()
-
--- a/chaco/base_2d_plot.py
+++ b/chaco/base_2d_plot.py
@@ -89,13 +89,13 @@ class Base2DPlot(AbstractPlotRenderer):
         self.trait_set(**kwargs_tmp)
         super(Base2DPlot, self).__init__(**kwargs)
         if self.index is not None:
-            self.index.on_trait_change(self._update_index_data,
+            self.index.observe(self._update_index_data,
                                        "data_changed")
         if self.index_mapper:
-            self.index_mapper.on_trait_change(self._update_index_mapper,
+            self.index_mapper.observe(self._update_index_mapper,
                                               "updated")
         if self.value is not None:
-            self.value.on_trait_change(self._update_value_data,
+            self.value.observe(self._update_value_data,
                                        "data_changed")
         # If we are not resizable, we will not get a bounds update upon layout,
         # so we have to manually update our mappers
@@ -247,7 +247,7 @@ class Base2DPlot(AbstractPlotRenderer):
     # Private methods
     #------------------------------------------------------------------------
 
-    def _update_index_mapper(self):
+    def _update_index_mapper(self, event=None):
         """ Updates the index mapper.
 
         Called by various trait change handlers.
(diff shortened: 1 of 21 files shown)


chaco/colormapped_selection_overlay.py
```diff
@@ -102,24 +103,43 @@ def overlay(self, component, gc, view_bounds=None, mode="normal"):
...
-            self.datasource_change_handler(new, "color_data", None, new.color_data)
+            
+            self.datasource_change_handler(
+                TraitChangeEvent(
+                    object=new,
+                    name="color_data",
+                    old=None,
+                    new=new.color_data
+                )
+            )
```
aaronayres35, line 120: This is certainly a pattern we would like to avoid.  For now I just went with the most straightforward transition possible to preserve behavior, but at some point we should rethink this

chaco/base_2d_plot.py
```diff
@@ -89,13 +89,13 @@ def __init__(self, **kwargs):
         self.trait_set(**kwargs_tmp)
         super(Base2DPlot, self).__init__(**kwargs)
         if self.index is not None:
-            self.index.on_trait_change(self._update_index_data,
+            self.index.observe(self._update_index_data,
                                        "data_changed")
```
rahulporuri, line 93:
```suggestion
            self.index.observe(self._update_index_data, "data_changed")
```

chaco/cmap_image_plot.py
```diff
@@ -171,10 +171,10 @@ def _set_color_mapper(self, val):
 
     def _value_mapper_changed(self, old, new):
         if old is not None:
-            old.on_trait_change(self._update_value_mapper,
+            old.observe(self._update_value_mapper,
                                 "updated", remove=True)
```
rahulporuri, line 175:
```suggestion
            old.observe(self._update_value_mapper, "updated", remove=True)
```

chaco/colormapped_scatterplot.py
```diff
@@ -394,10 +394,10 @@ def draw_func(x, y, size):
 
     def _color_data_changed(self, old, new):
         if old is not None:
-            old.on_trait_change(self._either_data_changed, "data_changed", remove=True)
+            old.observe(self._either_data_updated, "data_changed", remove=True)
         if new is not None:
-            new.on_trait_change(self._either_data_changed, "data_changed")
-        self._either_data_changed()
+            new.observe(self._either_data_updated, "data_changed")
+        self._either_data_updated()
```
rahulporuri, line 400: Hmm. I guess this method was inherited by this class? Renaming a method which is used by inherited classes can cause breakage. No changes necessary - it would be useful if we could search for users of the old method though.

chaco/colormapped_selection_overlay.py
```diff
@@ -102,24 +103,43 @@ def overlay(self, component, gc, view_bounds=None, mode="normal"):
...
+                    object=new,
+                    name="metadata_changed",
+                    old=None,
+                    new=new.metadata
+                )
+        )
         return
 
-    def selection_change_handler(self, obj, name, old, new):
+    def selection_change_handler(self, event):
```
rahulporuri, line 140: same here about modifying the API of a public method

chaco/colormapped_selection_overlay.py
```diff
@@ -102,24 +103,43 @@ def overlay(self, component, gc, view_bounds=None, mode="normal"):
...
+                    object=new,
+                    name="color_data",
+                    old=None,
+                    new=new.color_data
+                )
+            )
         return
 
-    def datasource_change_handler(self, obj, name, old, new):
+    def datasource_change_handler(self, event):
```
rahulporuri, line 123: sigh. there's no way around modifying the api of this public method. i'm not sure why this is a public method in the first place!

chaco/data_label.py
```diff
@@ -535,11 +535,11 @@ def _component_changed(self, old, new):
 
     def _modify_mapper_listeners(self, mapper, attach=True):
         if mapper is not None:
-            mapper.on_trait_change(self._handle_mapper, 'updated',
+            mapper.observe(self._handle_mapper, 'updated',
                                    remove=not attach)
```
rahulporuri, line 539:
```suggestion
            mapper.observe(self._handle_mapper, 'updated', remove=not attach)
```

chaco/grid.py
```diff
@@ -355,13 +355,13 @@ def _draw_component(self, gc, view_bounds=None, mode="normal"):
...
-            old.on_trait_change(self.mapper_updated, "updated", remove=True)
+            old.observe(self.mapper_updated, "updated", remove=True)
         if new is not None:
-            new.on_trait_change(self.mapper_updated, "updated")
+            new.observe(self.mapper_updated, "updated")
         self.invalidate()
         return
 
-    def mapper_updated(self):
+    def mapper_updated(self, event):
```
rahulporuri, line 364:
```suggestion
    def mapper_updated(self, event=None):
```

chaco/overlays/databox.py
```diff
@@ -160,7 +158,7 @@ def _update_bounds(self):
         self._data_bounds = [data_x2 - data_pos[0], data_y2 - data_pos[1]]
         self.trait_property_changed("data_bounds", self._data_bounds)
 
-    def my_component_moved(self):
+    def my_component_moved(self, event):
```
rahulporuri, line 161:
```suggestion
    def my_component_moved(self, event=None):
```

chaco/overlays/databox.py
```diff
@@ -169,7 +167,7 @@ def my_component_moved(self):
         self._bounds_valid = False
         self._position_valid = False
 
-    def my_component_resized(self):
+    def my_component_resized(self, event):
```
rahulporuri, line 170:
```suggestion
    def my_component_resized(self, event=None):
```

chaco/plot.py
```diff
@@ -1244,24 +1244,26 @@ def _color_mapper_changed(self):
 
     def _data_changed(self, old, new):
         if old:
-            old.on_trait_change(self._data_update_handler, "data_changed",
+            old.observe(self._data_update_handler, "data_changed",
                                 remove=True)
```
rahulporuri, line 1248:
```suggestion
            old.observe(self._data_update_handler, "data_changed", remove=True)
```

chaco/plotscrollbar.py
```diff
@@ -152,21 +152,17 @@ def _modify_plot_listeners(self, plot, action="attach"):
...
-        plot.on_trait_change(self._component_bounds_handler,
-                             "bounds_items", remove=remove)
-        plot.on_trait_change(self._component_pos_handler,
-                             "position", remove=remove)
-        plot.on_trait_change(self._component_pos_handler,
-                             "position_items", remove=remove)
+        plot.observe(self._component_bounds_handler,
+                             "bounds.items", remove=remove)
+        plot.observe(self._component_pos_handler,
+                             "position.items", remove=remove)
```
rahulporuri, line 158:
```suggestion
        plot.observe(self._component_bounds_handler,
                     "bounds.items", remove=remove)
        plot.observe(self._component_pos_handler,
                     "position.items", remove=remove)
```

chaco/scatterplot.py
```diff
@@ -526,7 +526,7 @@ def _outline_color_changed(self):
         self.invalidate_draw()
         self.request_redraw()
 
-    def _either_metadata_changed(self):
+    def _either_metadata_updated(self, event):
```
rahulporuri, line 529: just to confirm, this method was inherited?

rahulporuri (approved):
Mostly LGTM with a few suggestions

chaco/colormapped_scatterplot.py
aaronayres35: I did a search on github and there are some users. In many cases though it appears as though code was copy pasted along with the original method so this name change would not affect those users.  However, that is not always the case and some users who subclass like BaseXYPlot or ScatterPlot may use the inherited method via the old name

chaco/scatterplot.py
aaronayres35: Yep, from BaseXYPlot

chaco/colormapped_scatterplot.py
aaronayres35: @rahulporuri should we undo the name change here to be safe? Or do you think it is okay to go ahead with the change

rahulporuri: no i think it's fine, let's go ahead.

markasbach:
If anyone sees this: A typo slipped through with this pull request. I've opened an issue for this: https://github.com/enthought/chaco/issues/914

PR #591, opened 2021-04-02
Replace dynamic on_trait_change with observe part 2

aaronayres35:
This PR continues replacing the dynamic uses of on_trait_change with observe equivalents on top of #590

### DIFF
--- a/chaco/scatter_inspector_overlay.py
+++ b/chaco/scatter_inspector_overlay.py
@@ -7,6 +7,7 @@ from numpy import array, asarray
 # Enthought library imports
 from enable.api import ColorTrait, MarkerTrait
 from traits.api import Float, Int, Str, Trait
+from traits.observation.events import TraitChangeEvent
 
 # Local, relative imports
 from .abstract_overlay import AbstractOverlay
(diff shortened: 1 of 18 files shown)


chaco/tools/range_selection.py
```diff
@@ -679,15 +679,15 @@ def _determine_axis(self):
...
         return
 
     def _axis_changed(self, old, new):
         if old is not None:
-            self.plot.on_trait_change(self.__mapper_changed,
+            self.plot.observe(self.__mapper_changed,
                                       old + "_mapper", remove=True)
         if new is not None:
-            self.plot.on_trait_change(self.__mapper_changed,
-                                      old + "_mapper", remove=True)
```
aaronayres35, line 692: This previous code seems like it was a copy paste error...
Butt the updated code is still not right ad we have a case in the test suite where old is non None, but its listener never was hooked up initially

chaco/scatter_inspector_overlay.py
```diff
@@ -41,7 +42,7 @@ class ScatterInspectorOverlay(AbstractOverlay):
     # using a series of trait change handlers (defined at the end of the
     # class)
```
rahulporuri, line 43: This is tangential but i want to dig a little and find the source of this comment - it looks like there was going to be some fancy traits feature which would remove unnecessary code.

chaco/scatter_inspector_overlay.py
```diff
@@ -41,7 +42,7 @@ class ScatterInspectorOverlay(AbstractOverlay):
...
     #@on_trait_change('component.index.metadata_changed,component.value.metadata_changed')
-    def metadata_changed(self, object, name, old, new):
+    def metadata_updated(self, event):
```
rahulporuri, line 45: public API - just being safe

```suggestion
    def metadata_updated(self, event=None):
```

rahulporuri (approved):
LGTM with a couple of small comments.

There are a bunch of indentation errors that i'm ignoring because we'll run black on chaco at somepoint this month, which should fix all of those issues.

PR #592, opened 2021-04-02
Remove dummy return statements in testsuite

rahulporuri:
This PR removes dummy/unnecessary `return` statements in the testsuite.


aaronayres35 (approved):
LGTM

PR #593, opened 2021-04-02
Remove dummy return statements in docs and examples

rahulporuri:
This PR removes dummy/unnecessary `return` statements from the documentation and examples.


aaronayres35 (approved):
LGTM

PR #594, opened 2021-04-02
Remove unnecessary return statements in package

rahulporuri:
This PR removes unnecessary `return` statements from the `chaco` package.

### DIFF
--- a/chaco/multi_line_plot.py
+++ b/chaco/multi_line_plot.py
@@ -374,8 +374,6 @@ class MultiLinePlot(BaseXYPlot):
 
         self._cached_data_pts = line_points
         self._cache_valid = True
-        return
-
 
     # See base_xy_plot.py for:
     ## def _downsample(self):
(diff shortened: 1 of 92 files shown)


chaco/multi_line_plot.py
```diff
@@ -443,28 +439,23 @@ def _render_icon(self, gc, x, y, width, height):
...
 
     def _line_style_changed(self):
         self.invalidate_draw()
         self.request_redraw()
-        return
 
     def _line_width_changed(self):
         self.invalidate_draw()
         self.request_redraw()
-        return
```
aaronayres35, line 461: unrelated to this PR but looks like these could all use the new `requires_redraw` metadata

aaronayres35 (approved):
LGTM

chaco/multi_line_plot.py
rahulporuri: can you please create a general purpose issue to search and see where all we could use `requires_redraw` in chaco?

aaronayres35: I've opened #595

Issue #583, opened 2021-03-25
Audit __getstate__ and __setstate__ methods


corranwebster:
There are a number of Chaco classes which override `__getstate__` and possibly `__setstate__`.  Most of these seem to be removing traits from the state dict, which is now correctly handled by marking those traits as `transient`.

For example:

https://github.com/enthought/chaco/blob/e3423614d654a203bd82ee087c085abe13b465b1/chaco/lineplot.py#L415-L421

This is removing the `traits_view` trait from the state - under normal Traits, this will never be in the state because of special handling of Views, so this whole method can be removed.

Another example:

https://github.com/enthought/chaco/blob/e3423614d654a203bd82ee087c085abe13b465b1/chaco/grid.py#L413-L419

Here the traits `_cache_valid`, `_tick_list`, `_tick_positions`, `_tick_extents` are being removed from the state.  These traits should be marked with `transient=True` metadata and then the `__getstate__` method can be removed.

If it turns out that there are necessary things being done in a `__getstate__`, we should make sure that there is a `super` call.

PR #598, opened 2021-04-02
Remove serializable module and its tests

### LINKED ISSUE #250: Serializable mixin should be removed

aaronayres35:
closes #250

This PR removes the Serializable mixin class and related tests.

Supersedes #575.  We will deal with `_post_load` separately as per this comment: https://github.com/enthought/chaco/pull/582#issuecomment-806514696

also there is #583 for dealing with __get_state__ and __set_state__

To keep things simple here we just remove serializable and its tests


rahulporuri:
Can you resolve the merge conflicts? Looks like the black changes conflict with the removals in this PR.

rahulporuri (approved):
LGTM, barring the merge conflicts.

PR #600, opened 2021-04-02
run flake8 on CI but ignore all existing errors

aaronayres35:
NOTE: this PR sits on top of #599

I figured having run black on the code base before setting up flake8 machinery made sense

This PR sets up flake8 to run on CI so that new flake8 errors aren't introduced.  It ignores all the currently existing flake8 errors (there are a lot, filenames were found via `flake8 -q` and added to the tox.ini exclude list)


tox.ini
```diff
@@ -0,0 +1,236 @@
+[flake8]
+per-file-ignores = */api.py:F401
+exclude =
+    # The following list should eventually be empty.
+    # list can be found with flake8 -q
+    setup.py
+    ci/edmtool.py
+    docs/source/conf.py
```
rahulporuri, line 8: these three files aren't targets in the `ci\edmtool flake8` command

rahulporuri (approved):
LGTM

Issue #596, opened 2021-04-02
Clarify / Remove use of metadata, metadata_changed, etc traits


aaronayres35:
**Problem Description**
Currently Chaco does some pretty confusing things in regards to metadata.
There are:
 - traits named metadata
 - traits name metadata_changed
 - magically named `_metadata_changed` methods which simply set the `metadata_changed` trait to `True` when the `metadata` trait changes
 - a method name `metadata_changed` (no leading underscore) that gets hooked up as a change handler listening for changes on the `metadata_changed` trait. (Note this method has been renamed to `metadata_updated`)

All of this is rather confusing, and likely unnecessary.
We need to investigate how all of this is working more closely and see if it can all be replaced with newer traits features.

aaronayres35:
ref: #398

Issue #398, opened 2018-01-29
Deprecate metadata_changed event


jvkersch:
With the merge of https://github.com/enthought/chaco/pull/395, `ScatterInspector.metadata_changed` is redundant, and `ScatterInspector.inspector_event` should be used instead. This issue is to document:

- [ ] Switching over uses in the codebase of the former to the latter, and putting some more warnings in place that `metadata_changed` is redundant.
- [ ] Removing the actual event after a couple of releases have gone by.

PR #603, opened 2021-04-02
Remove metadata_changed event trait

aaronayres35:
part of #596 

This PR incrementally removes the use of the `metadata_changed` event trait defined on some of the data source classes.  This trait was (_almost_) only ever set via the `_metadata_changed` and `_metadata_items_changed` magically named static handlers.  As a result, functionality could be preserved by simply directly listening to `metadata.items` in all places we had previously listened to `metadata_changed`.  The few cases where other things were done with the event (ie we explicitly set `metadata_changed` to something) have also been dealt with as needed on a case by case basis.

I tried to make these changes incrementally and the test suite passed locally on every commit. 

This is hopefully a first step towards resolving #596, but there is likely more that can be done to clean up this stuff / use the latest traits capabilites.


examples/demo/financial/correlations.py
```diff
@@ -169,16 +169,13 @@ def _create_data(self, *names):
         self.plotdata = plotdata
 
     def _selections_updated(self, event):
-        metadata_changed_event = event.new
+        new_metadata = event.object
         if self.corr_renderer is None:
             return
-        if (
-            not isinstance(metadata_changed_event, dict)
```
aaronayres35, line 176: We previously checked if the event was a dict, but now we replace `.` with `:`, ie we listen to `metadata:items` so that this check is no longer needed

chaco/colormapped_selection_overlay.py
```diff
@@ -119,25 +119,25 @@ def _component_changed(self, old, new):
             )
 
     def datasource_change_handler(self, event):
-        obj, name, old, new = (event.object, event.name, event.old, event.new)
```
aaronayres35, line 122: a bit orthogonal to this PR, but this name was unused

chaco/tools/range_selection.py
```diff
@@ -591,7 +591,6 @@ def _set_selection(self, val):
 
             # Set the selection range on the datasource
             datasource.metadata[mdname] = val
-            datasource.metadata_changed = {mdname: val}
```
aaronayres35, line 594: it appears as though the `metadata` dict is being changed only once here, however it is being set potentially multiple times below.

Perhaps we only want the one change event to occur rather than being notified for all the changes.  This is the only place I am explicitly seeing where there may be a change in behavior from this.  If this does indeed change behavior, we will likely want to revert these removals.  

Perhaps this PR should simply be closed as `metadata_changed` seems here to stay.  I still do find it confusing and in any case we should more clearly document _why_ it exists, and how it is meant to be used differently than just listening to changes on the metadata directly (if that is in fact true)

rahulporuri (commented):
I thought about this some more - and I think we should leave `metadata_changed` as it is i.e. don't try to remove it - and don't try to remove uses of it.

It actually is a nice API that the abstract data source implementers provide - and the plot renderers don't have to worry about the specific trait type of `metadata` trait.

I apogolize for not looking at this issue more closely earlier - and for letting this PR/issue sit idle for this long.

Does this make sense to you?

aaronayres35:
> I thought about this some more - and I think we should leave `metadata_changed` as it is i.e. don't try to remove it - and don't try to remove uses of it.
> 
> It actually is a nice API that the abstract data source implementers provide - and the plot renderers don't have to worry about the specific trait type of `metadata` trait.
> 
> I apogolize for not looking at this issue more closely earlier - and for letting this PR/issue sit idle for this long.
> 
> Does this make sense to you?

Yeah that makes sense, I will close this PR

aaronayres35:
> I thought about this some more - and I think we should leave `metadata_changed` as it is i.e. don't try to remove it - and don't try to remove uses of it.
> 
> It actually is a nice API that the abstract data source implementers provide - and the plot renderers don't have to worry about the specific trait type of `metadata` trait.
> 
> I apogolize for not looking at this issue more closely earlier - and for letting this PR/issue sit idle for this long.
> 
> Does this make sense to you?

I had forgotten about #398, it seems that at least for `ScatterInspector` `inspector_event` can be used instead.  In general though I agree it should stay

PR #625, opened 2021-04-05
Replace some _anytrait_changed methods

### LINKED ISSUE #624: Deal with _anytrait_changed handlers
**Problem Description**
In many cases in chaco we have `_anytrait_changed` methods which define a list of traits of actual interest, checks if the changed trait is in that list, and then calls a certain method.  For example:
https://github.com/enthought/chaco/blob/c758a65222a2caef9a3d2c074ae9dd5775a3a77e/chaco/axis.py#L781-L816

This is a case where we would want to use metadata in the trait definition and listen based on that.  Similar to the recently added `requires_redraw` (see PR #585 and issue #595)

aaronayres35:
closes #624 

Note there are still a few examples using `_anytrait_changed` methods that can be addressed in a different PR.  I ended up not using metadata in the `legend.py` and just went with the straight forward conversion to observe.  We can come back to this if we want to be a little more clever later.

### DIFF
--- a/chaco/legend.py
+++ b/chaco/legend.py
@@ -503,11 +503,11 @@ class Legend(AbstractOverlay):
         "bounds.items",
         "title_at_top",
     ])
-    def _update_layout_needed(self, event):
+    def _invalidate_existing_layout(self, event):
         self._layout_needed = True
     
     @observe("color")
-    def _get_preferred_size_on_color_change(self, event):
+    def _update_caches(self, event):
         self.get_preferred_size()
 
     def _plots_changed(self):
--- a/chaco/legend.py
+++ b/chaco/legend.py
@@ -503,7 +503,7 @@ class Legend(AbstractOverlay):
         "bounds.items",
         "title_at_top",
     ])
-    def _udpate_layout_needed(self, event):
+    def _update_layout_needed(self, event):
         self._layout_needed = True
     
     @observe("color")
--- a/chaco/legend.py
+++ b/chaco/legend.py
@@ -19,6 +19,7 @@ from traits.api import (
     Instance,
     Int,
     List,
+    observe,
     Str,
 )
 
@@ -487,29 +488,27 @@ class Legend(AbstractOverlay):
         return CompositeIconRenderer()
 
     # -- trait handlers --------------------------------------------------------
-    def _anytrait_changed(self, name, old, new):
-        if name in (
-            "font",
-            "border_padding",
-            "padding",
-            "line_spacing",
-            "icon_bounds",
-            "icon_spacing",
-            "labels",
-            "plots",
-            "plots_items",
-            "labels_items",
-            "border_width",
-            "align",
-            "position",
-            "position_items",
-            "bounds",
-            "bounds_items",
-            "label_at_top",
-        ):
-            self._layout_needed = True
-        if name == "color":
-            self.get_preferred_size()
+    @observe([
+        "font",
+        "border_padding",
+        "padding",
+        "line_spacing",
+        "icon_bounds",
+        "icon_spacing",
+        "labels.items",
+        "plots.items",
+        "border_width",
+        "align",
+        "position.items",
+        "bounds.items",
+        "title_at_top",
+    ])
+    def _udpate_layout_needed(self, event):
+        self._layout_needed = True
+    
+    @observe("color")
+    def _get_preferred_size_on_color_change(self, event):
+        self.get_preferred_size()
 
     def _plots_changed(self):
         """Invalidate the caches."""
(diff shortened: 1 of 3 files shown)


aaronayres35:
The occurrence in `chaco/legend.py` is a bit trickier as there we check for `_items` traits.  I am unsure on the cleanest way to do this using metadata like we do for the `axis.py` case.  It may be that we just add those things explicitly in the expression passed to observe (e.g `observe("plots.items, labels.items, +needs_layout")` or something like this)

However, this may lead to some confusion.  Additionally, the metadata use in `chaco/axis.py` may already be a little confusing.  For example the `title` trait does not have `is_visual` set to `True`, because currently it is handled differently with its own change handler (we effectively do the same as calling `_invalidate` but we do not set `self._cache_valid = False`).  Likewise there is a `bgcolor` trait that currently does not have `is_visual` set because it was not previously listed in the `_anytrait_changed` method, but this certainly seems to be a visual attribute.  Perhaps it should trigger `_invalidate` when it changes?  Or perhaps there was some reason for it not being included?

aaronayres35:
looks like the intermittent failure on `test_dont_crash_on_click` is not windows specific... 😞 

On linux/pyqt
```
ERROR: test_dont_crash_on_click (chaco.tests.test_plot.TestEmptyPlot)
  File "/home/travis/.edm/envs/chaco-test-3.6-pyqt/lib/python3.6/site-packages/chaco/tests/test_plot.py", line 148, in test_dont_crash_on_click
    [(1, 1), (2, 2), (3, 3), (4, 4)],
  File "/home/travis/.edm/envs/chaco-test-3.6-pyqt/lib/python3.6/contextlib.py", line 88, in __exit__
    next(self.gen)
  File "/home/travis/.edm/envs/chaco-test-3.6-pyqt/lib/python3.6/site-packages/traitsui/testing/tester/ui_tester.py", line 106, in create_ui
    process_cascade_events()
  File "/home/travis/.edm/envs/chaco-test-3.6-pyqt/lib/python3.6/contextlib.py", line 88, in __exit__
    next(self.gen)
  File "/home/travis/.edm/envs/chaco-test-3.6-pyqt/lib/python3.6/site-packages/traitsui/testing/_exception_handling.py", line 74, in reraise_exceptions
    raise RuntimeError(msg)
RuntimeError: Uncaught exceptions found.
=== Exception (type: <class 'AttributeError'>, value: '_QtWindow' object has no attribute 'handler') ===
  File "/home/travis/.edm/envs/chaco-test-3.6-pyqt/lib/python3.6/site-packages/enable/qt4/base_window.py", line 255, in enterEvent
    self.handler.enterEvent(event)
AttributeError: '_QtWindow' object has no attribute 'handler'
=== Exception (type: <class 'AttributeError'>, value: '_QtWindow' object has no attribute 'handler') ===
  File "/home/travis/.edm/envs/chaco-test-3.6-pyqt/lib/python3.6/site-packages/enable/qt4/base_window.py", line 264, in mouseMoveEvent
    self.handler.mouseMoveEvent(event)
AttributeError: '_QtWindow' object has no attribute 'handler'
=== Exception (type: <class 'AttributeError'>, value: '_QtWindow' object has no attribute 'handler') ===
  File "/home/travis/.edm/envs/chaco-test-3.6-pyqt/lib/python3.6/site-packages/enable/qt4/base_window.py", line 264, in mouseMoveEvent
    self.handler.mouseMoveEvent(event)
AttributeError: '_QtWindow' object has no attribute 'handler'
=== Exception (type: <class 'AttributeError'>, value: '_QtWindow' object has no attribute 'handler') ===
  File "/home/travis/.edm/envs/chaco-test-3.6-pyqt/lib/python3.6/site-packages/enable/qt4/base_window.py", line 258, in leaveEvent
    self.handler.leaveEvent(event)
AttributeError: '_QtWindow' object has no attribute 'handler'
```

chaco/legend.py
```diff
@@ -487,29 +488,27 @@ def _composite_icon_renderer_default(self):
...
+        "line_spacing",
+        "icon_bounds",
+        "icon_spacing",
+        "labels.items",
+        "plots.items",
+        "border_width",
+        "align",
+        "position.items",
+        "bounds.items",
+        "title_at_top",
```
rahulporuri, line 504:
```suggestion
        "label_at_top",
```

rahulporuri (approved):
LGTM

chaco/legend.py
aaronayres35: I believe this had actually previously been a bug. There is no `label_at_top` trait defined on this class, but there is a `title_at_top`

rahulporuri: ahh. never mind then. please go ahead.

Issue #595, opened 2021-04-02
Look for places to use new requires_redraw metadata


aaronayres35:
**Problem Description**
In PR #585 we added the `requires_redraw` metadata to specify that if a particular trait changes, `invalidate_and_redraw` should be called.  We should sweep the code bases for all possible cases where we could use this, eg for things like:
```
def _sometrait_change(self):
    self.invalidate_draw()
    self.request_redraw()
```

PR #626, opened 2021-04-05
Use requires_redraw when possible

aaronayres35:
part of #595 

This PR uses the new `requires_redraw` metadata on a few `PlotComponent` subclasses.

This covers the simplest cases for where `requires_redraw` can be used.  There are other scenarios where it may be useful but would require doing something a bit differently.  I will leave those decisions / changes for a later PR


chaco/lineplot.py
```diff
@@ -433,22 +433,6 @@ def _downsample_vectorized(self):
         d = z[:, 0] + z[:, 1]
         # ... TODO ...
 
-    def _alpha_changed(self):
-        self.invalidate_draw()
-        self.request_redraw()
-
```
aaronayres35, line 439: Note `alpha` has `required_redraw=True` on he base class `BaseXYPlot` from #585 

See comment: https://github.com/enthought/chaco/pull/585#discussion_r605437118

rahulporuri (approved):
LGTM.

PR #631, opened 2021-04-07
Add GitHub Actions CI workflow

### LINKED ISSUE #614: Use GitHub Actions instead of Travis and Appveyor  CI
We need to replace the use of Travis and Appveyor CI with GitHub Actions CI, similar to what we did in other ETS projects.

aaronayres35:
closes #614 

This PR simply adds GH actions workflow to run CI (Following example of enable, and doing what we currently do on travis/appveyor)
For now, as with other ets projects, this PR adds gh actions but does not yet remove appveyor or travis


aaronayres35:
Also, orthogonal to this PR, but in other ets projects we have a check for if the toolkit specified is Pyside2 and we then pip install it (as it is not available through free edm).  It looks like chaco doesn't do this though?  It instead adds an additional edm repository to get it through edm.  Is this the correct approach or should this be changed @rahulporuri

aaronayres35:
Looks like we need to enable Actions in the repo settings?

aaronayres35:
Im unsure why gh actions flake8 command was picking up errors travis was not.  Nonetheless, CI is green now with GH actions running

rahulporuri (approved):
LGTM

PR #632, opened 2021-04-07
[WIP] Contribute examples to etsdemo

aaronayres35:
I did not do a great job of making these changes commit by commit.
In any case, so far this PR sets up the `entry_point`, moves the examples in `examples/demo` into `chaco/examples/demo`,  adds an `__extras_require__` to setup.py so that example dependencies can be easily installed via pip (although pip installing chaco/enable can still be a pain).

To test, one can do the following:
```
python ci/edmtool.py install --toolkit=pyqt5
edm shell -e chaco-test-3.6-pyqt5
pip install -e .[examples]
pip install etsdemo
etsdemo
```

We still need to have a discussion where we thoroughly go through examples and decide which ones we want to contribute, which we may want to remove entirely, etc.  In that sense, this PR may be jumping the gun a little early (hence the current WIP title). Nonetheless, this gets everything all setup in prep for those decisions to be made


rahulporuri:
can you hold off on recreating this PR? I would prefer starting small instead of including the entire top-level examples directory into chaco/examples.

aaronayres35:
> can you hold off on recreating this PR? I would prefer starting small instead of including the entire top-level examples directory into chaco/examples.

yep will do

PR #633, opened 2021-04-07
Deal with __getstate__ and __setstate__ methods

### LINKED ISSUE #583: Audit __getstate__ and __setstate__ methods

aaronayres35:
closes #583 

This PR rermoves uses of the `__getstate__` method when it is not needed, and cleans up its use where it is.

The `__setstate__` methods can be dealt with in a follow up PR


chaco/abstract_mapper.py
```diff
@@ -49,13 +49,6 @@ def map_data_array(self, screen_vals):
     # ------------------------------------------------------------------------
     # Persistence-related methods
     # ------------------------------------------------------------------------
-    def __getstate__(self):
-        state = super(AbstractMapper, self).__getstate__()
-        for key in ["_cache_valid"]:
```
aaronayres35, line 54: `_cache_valid` is not a trait on this class or any of its base classes

chaco/tools/simple_zoom.py
```diff
@@ -70,25 +70,25 @@ class SimpleZoom(AbstractOverlay, ToolHistoryMixin, BaseZoomTool):
...
-    minimum_screen_delta = Int(10)
+    minimum_screen_delta = Int(10, transient=True)
 
     # -------------------------------------------------------------------------
     # Appearance properties (for Box mode)
     # -------------------------------------------------------------------------
 
     # The pointer to use when drawing a zoom box.
-    pointer = "magnifier"
+    pointer = Str("magnifier", transient=True)
```
aaronayres35, line 91: This may be bad if `pointer` is meant to be a class attribute, similar to `stack_index` in `plot_container.py`.

I need to climb up the class hierarchy to see if it was previously defined as a trait.  I remember Kit mentioning this being a point of confusion.  I am not sure what the ideal way to avoid this confusion would be (aside from maybe a comment above saying if this is a class attribute or just setting a trait to a value.  I guess if setting a trait to some value, you could do as I have done here? and explicitly do something like `Str(___)`

chaco/tests/test_arraydatasource.py
```diff
@@ -247,7 +254,7 @@ def test_serialization_state_no_persist(self):
             "_min_index",
             "_max_index",
         ]:
-            self.assertIn(key, state)
+            self.assertNotIn(key, state)
```
aaronayres35, line 257: If not persisting they should not be in state, but if we are they should be. I believe these tests are doing what they are supposed to now

chaco/tools/simple_zoom.py
aaronayres35: As a quick sanity check, on master I defined a class attribute `thing = "A"` on `SimpleZoom`.  I then ran:
```
from chaco.tools.simple_zoom import SimpleZoom
SimpleZoom.thing  # I see 'A' printed
SimpleZoom.pointer  # I get an error "type object 'SimpleZoom' has no attribute 'pointer'"
```

So, I believe the change made here is the right one

chaco/plot_containers.py
```diff
@@ -281,9 +281,8 @@ def _do_stack_layout(self, components, align):
     # PICKLE FIXME: blocked with _pickles, but not sure that was correct.
     def __getstate__(self):
         state = super(StackedPlotContainer, self).__getstate__()
-        for key in ["stack_dimension", "other_dimension", "stack_index"]:
-            if key in state:
-                del state[key]
+        if "stack_index" in state:
+            del state["stack_index"]
```
aaronayres35, line 285: stack_index is a class attribute not a trait, so I believe this is still needed

chaco/abstract_mapper.py
rahulporuri: sigh. but it is a trait on the subclasses e.g. `Base1DMapper`

chaco/lineplot.py
```diff
@@ -449,14 +449,6 @@ def _line_width_changed(self):
         self.invalidate_draw()
         self.request_redraw()
 
-    def __getstate__(self):
-        state = super(LinePlot, self).__getstate__()
-        for key in ["traits_view"]:
```
rahulporuri, line 454: Is `traits_view` a transient trait? I dont see anything in the code or in the documentation saying that it is.

chaco/tools/simple_zoom.py
```diff
@@ -647,28 +647,3 @@ def _next_state_pressed(self):
...
-        dont_pickle = [
-            "always_on",
-            "always_on_modifier",
-            "enter_zoom_key",
-            "exit_zoom_key",
-            "minimum_screen_delta",
-            "event_state",
-            "reset_zoom_key",
-            "prev_zoom_key",
-            "next_zoom_key",
```
rahulporuri, line 663: im not sure where these traits were being magically defined

rahulporuri (approved):
LGTM with a couple of comments

chaco/lineplot.py
aaronayres35: I was going off wha Corran said in this issue description: "This is removing the traits_view trait from the state - under normal Traits, this will never be in the state because of special handling of Views, so this whole method can be removed."
This should certainly be documented though

chaco/tools/simple_zoom.py
aaronayres35: me either, this is the only place I was seeing them in the codebase. Perhaps they once existed but had been removed

chaco/abstract_mapper.py
aaronayres35: Ah you are right.  TBH I believe that all related traits should be transient even if they weren't perviously involved in a `__getstate__`method.  For example other classes have `_screen_cache_valid` or `_selection_cache_valid` traits.  I think I will update all of these to be transient as I can't see why they shouldn't all be so

aaronayres35:
Looks like the CI failure on windows/pyqt from the test introduced by #636 is reproducible on every PR except the original (no idea how it didn't fail on 636).
```
ERROR: test_dont_crash_on_click (chaco.tests.test_plot.TestEmptyPlot)
  File "D:\a\chaco\chaco\.edm\envs\chaco-test-3.6-pyqt\lib\site-packages\chaco\tests\test_plot.py", line 134, in test_dont_crash_on_click
    [(1, 1), (25, 25), (50, 50), (100, 100)],
  File "D:\a\chaco\chaco\.edm\envs\chaco-test-3.6-pyqt\lib\contextlib.py", line 88, in __exit__
    next(self.gen)
  File "D:\a\chaco\chaco\.edm\envs\chaco-test-3.6-pyqt\lib\site-packages\traitsui\testing\tester\ui_tester.py", line 106, in create_ui
    process_cascade_events()
  File "D:\a\chaco\chaco\.edm\envs\chaco-test-3.6-pyqt\lib\contextlib.py", line 88, in __exit__
    next(self.gen)
  File "D:\a\chaco\chaco\.edm\envs\chaco-test-3.6-pyqt\lib\site-packages\traitsui\testing\_exception_handling.py", line 74, in reraise_exceptions
    raise RuntimeError(msg)
RuntimeError: Uncaught exceptions found.
=== Exception (type: <class 'AttributeError'>, value: '_QtWindow' object has no attribute 'handler') ===
  File "D:\a\chaco\chaco\.edm\envs\chaco-test-3.6-pyqt\lib\site-packages\enable\qt4\base_window.py", line 258, in leaveEvent
    self.handler.leaveEvent(event)
AttributeError: '_QtWindow' object has no attribute 'handler'
```

I also find the error quite confusing, as looking in enable, the `_QtWindow` class clearly has `self.handler = _QtWindowHandler(self, enable_window)` in its `__init__` method... Also this seems like a very strange thing to be windows specific... 🤔 
I will look into this more

chaco/base_contour_plot.py
```diff
@@ -54,10 +54,10 @@ class BaseContourPlot(Base2DPlot):
     # ------------------------------------------------------------------------
 
     # Is the cached level data valid?
-    _level_cache_valid = Bool(False)
+    _level_cache_valid = Bool(False, transient=True)
 
     # Is the cached color data valid?
-    _colors_cache_valid = Bool(False)
+    _colors_cache_valid = Bool(False, transient=True)
```
rahulporuri, line 60: If the cache valid flags arent transient, then the caches themselves should also be transient - `_levels` and `_colors`.

I would think `HasTraits` would ignore traits that are private i.e. that start an `_`. If that isn't the case, there are a lot more private traits that we will want to mark as transient.

aaronayres35: As you suggested, I printed `state` inside the `ArrayDataSource.__getstate__` method on master and ran the test suite.  I do see `_` prefixed traits, eg `_data`, '_min_index', etc.  So it looks like these need to be addressed

aaronayres35: >If that isn't the case, there are a lot more private traits that we will want to mark as transient.

Does this mean all private traits be transient?

rahulporuri: > As you suggested, I printed state inside the ArrayDataSource.__getstate__ method on master and ran the test suite. I do see _ prefixed traits, eg _data, '_min_index', etc. So it looks like these need to be addressed

Sigh. I was afraid of that.

> Does this mean all private traits be transient?

Let's not do that. Definitely not in this PR.

rahulporuri: So, I imagine the code will recreate the cache if a `*cache*` is `False`, which it is by default because we're marking it as a transient. So, I think we can ignore all of the traits that are recalculated based on the `*cache*` flag. - i.e. nothing actionable in this PR.

chaco/contour_line_plot.py
```diff
@@ -46,16 +46,16 @@ class ContourLinePlot(BaseContourPlot):
...
     # Is the cached width data valid?
-    _widths_cache_valid = Bool(False)
+    _widths_cache_valid = Bool(False, transient=True)
 
     # Is the cached style data valid?
-    _styles_cache_valid = Bool(False)
+    _styles_cache_valid = Bool(False, transient=True)
 
     # Cached list of line widths
     _widths = List
```
rahulporuri, line 61: looks like this is technically a cache but like I said earlier, this will be ignored/recomputed because the `_widths_cache_valid` trait is `False` - no action required, just pointing it out.

chaco/tools/simple_zoom.py
```diff
@@ -38,13 +38,23 @@ class SimpleZoom(AbstractOverlay, ToolHistoryMixin, BaseZoomTool):
...
 
+<<<<<<< HEAD
+    # Is the tool always "on"? If True, left-clicking always initiates
+    # a zoom operation; if False, the user must press a key to enter zoom mode.
+    always_on = Bool(False, transient=True)
+
+    # Defines a meta-key, that works with always_on to set the zoom mode. This
+    # is useful when the zoom tool is used in conjunction with the pan tool.
+    always_on_modifier = Enum(None, "shift", "control", "alt", transient=True)
+=======
```
rahulporuri, line 49: looks like a merge conflict crept in - and the chaco testsuite sadly didnt capture this!

chaco/tools/simple_zoom.py
```diff
@@ -68,27 +78,48 @@ class SimpleZoom(AbstractOverlay, ToolHistoryMixin, BaseZoomTool):
     #: Conversion ratio from wheel steps to zoom factors.
     wheel_zoom_step = Float(1.0)
 
+<<<<<<< HEAD
```
rahulporuri, line 81: another merge conflict

rahulporuri: looks like there are a few more in this module

aaronayres35: Oh yikes! 😓 thank you for catching these!

chaco/tools/simple_zoom.py
```diff
@@ -38,23 +38,13 @@ class SimpleZoom(AbstractOverlay, ToolHistoryMixin, BaseZoomTool):
     #:   Perform a "box" selection on two axes.
     tool_mode = Enum("box", "range")
 
-<<<<<<< HEAD
     # Is the tool always "on"? If True, left-clicking always initiates
     # a zoom operation; if False, the user must press a key to enter zoom mode.
```
rahulporuri: looks like you lost the `#:` because of the merge conflicts.

rahulporuri (approved):
LGTM

PR #634, opened 2021-04-08
Dont use traitsui.api as tui

### LINKED ISSUE #611: Dont use traitsui.api as tui
https://github.com/enthought/chaco/blob/3de7780561fa29e79c887432d3ce408ea82d1614/chaco/plugin/plot_editor.py makes use of the odd `import traitsui.api as tui` alias which needs to be updated and removed.

aaronayres35:
fixes #611 

This PR simply imports the needed objects from `traitsui.api` rather than importing the whole thing as tui.

As a side note, this module is based around Workbench which has been deprecated in favor of tasks.  Perhaps we want to rewrite this module? Or deprecate, or remove it... 🤔


rahulporuri (approved):
LGTM. I'm on board with removing it entirely but let's do that in a separate PR.

PR #637, opened 2021-04-08
Replace handmade api docs with autogenerated api docs

aaronayres35:
This PR replaces the hand curated api docs in the api reference with the autogenerated api docs.  It then deletes the files for the hand made api docs.

This PR _drastically_ decreases the number of warnings issued when building the docs from 1000+ down to now ~30.

In a follow up PR (possibly to sit on top of this branch), I will take the lists of links to important objects from these remove hand curated api doc files, and add those lists to the corresponding section of the user manual.  Having these clear listings is useful but we don't need the redundancy.

UPDATE:
This PR now uses api docs templates following the lead of PRs https://github.com/enthought/envisage/pull/339 and https://github.com/enthought/traits-futures/pull/181


aaronayres35:
One thing this PR has made me notice is that some modules in chaco explicitly define `__all__`, eg https://github.com/enthought/chaco/blob/fdd858aa6dbc76addb50d011fb81e879ce8e0355/chaco/plot_containers.py#L45-L50

As a result, in this particular case, the base class `StackedPlotContainer` does not show up in the auto generated api docs.  I believe this class shouldn't be used directly, rather is subclasses `HPlotContainer`/`VPlotContainer` should.  However, having it documented in the api docs could be useful.

aaronayres35:
@rahulporuri I am going to hold up on the follow up PR for now as I want to discuss with you more.  We had said offline that we could list these classes in their corresponding pages of he usermanual which makes sense, but what about the corresponding section of the Commonly used Modules and Classes page, as that appears to be sort of the point of that page.

rahulporuri (approved):
The autogenerated docs themselves look good but there are a couple of comments : 

- can we use the api docs templates (like we do in e.g. traits-futures)
- can we also ignore the `chaco.tests_with_backend` module when autogenerating the docs?

ci/edmtool.py
```diff
@@ -305,13 +307,23 @@ def docs(runtime, toolkit, environment):
...
 
+    api_path = os.path.join("docs", "source", "api")
+
+    # Remove any previously autogenerated API documentation.
+    doc_api_files = os.listdir(api_path)
+    permanent = ['chaco.api.rst', 'templates']
+    previously_autogenerated = \
+        [file for file in doc_api_files if file not in permanent]
+    for file in previously_autogenerated:
+        os.remove(os.path.join(api_path, file))
```
rahulporuri, line 318: This might be overkill but okay (not an actionable comment). I guess this is needed if we rename modules and dont do git clean locally between docs builds.

docs/source/api/chaco.api.rst
```diff
@@ -0,0 +1,13 @@
+..
+   (C) Copyright 2021 Enthought, Inc., Austin, TX
```
rahulporuri, line 2: nit. the year range needs to be fixed

docs/source/api/templates/module.rst_t
```diff
@@ -0,0 +1,21 @@
+..
+   (C) Copyright 2021 Enthought, Inc., Austin, TX
```
rahulporuri, line 2: nit - year range again

docs/source/api/templates/package.rst_t
```diff
@@ -0,0 +1,38 @@
+..
+   (C) Copyright 2021 Enthought, Inc., Austin, TX
```
rahulporuri, line 2: nit - year range again

rahulporuri (approved):
Still LGTM with a couple of comments

docs/source/api/chaco.api.rst
aaronayres35: 👍 
I saw 2006-2021 in the `LICENSE.txt` file so I assume 2006 is the correct start year.

PR #644, opened 2021-04-12
Import version from installed package if the file doesnt exist

rahulporuri:
If the package is installed in a non-editable way, the `_version` file won't exist where we expect it to. This PR updates the sphinx configuration to get the version information from the package directly if the file doesn't exist.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -58,10 +58,10 @@ if os.path.isfile(version_file):
     version = release = runpy.run_path(version_file)['full_version']
 else:
     try:
-        from enable._version import full_version as version
+        from chaco._version import full_version as version
         release = version
     except ImportError:
-        raise RuntimeError("Enable must be installed before building docs!")
+        raise RuntimeError("Chaco must be installed before building docs!")
 
 # There are two options for replacing |today|: either, you set today to some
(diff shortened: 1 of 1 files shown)


docs/source/conf.py
```diff
@@ -52,11 +52,16 @@ def get_build_docset():
...
+base_path = os.path.abspath(os.path.dirname(__file__))
+version_file = os.path.join(base_path, '..', '..', 'chaco', '_version.py')
+if os.path.isfile(version_file):
+    version = release = runpy.run_path(version_file)['full_version']
+else:
+    try:
+        from enable._version import full_version as version
+        release = version
+    except ImportError:
+        raise RuntimeError("Enable must be installed before building docs!")
```
aaronayres35, line 64: Looks like 2 copy paste errors from `enable`.

```suggestion
        from chaco._version import full_version as version
        release = version
    except ImportError:
        raise RuntimeError("Chaco must be installed before building docs!")
```

aaronayres35 (approved):
LGTM

just one comment about using `chaco` instead of `enable`.

rahulporuri:
sigh. thanks for catching that. silly _silly_ mistake

PR #645, opened 2021-04-12
Remove examples for the vtk backend in chaco

### LINKED ISSUE #608: Remove chaco examples for the enable vtk backend
The examples https://github.com/enthought/chaco/tree/master/examples/demo/vtk and https://github.com/enthought/chaco/blob/master/examples/demo/vtk_example.py use the now non-existent vtk backend in `enable` so they should be removed.

rahulporuri:
fixes #608 

The vtk backend was removed in version 5.0.0 of enable so these examples dont work anymore.


aaronayres35 (approved):
LGTM

PR #647, opened 2021-04-12
Remove miscellaneous/dead code

rahulporuri:
This PR simply removes unnecessary coding cookies in files and an end of file marker we missed in an earlier PR.


aaronayres35 (approved):
LGTM

Issue #640, opened 2021-04-09
Are ChacoPlotItem, ChacoEditorFactory, ChacoPlotEditor still relevant given ComponentEditor?


rahulporuri:
Chaco provides a TraitsUI Item, Editor and Editor Factory for a Chaco Plot - but given the presence of the enable `ComponentEditor`, is this still relevant or should these classes be deprecated/removed. What functionality do these classes provide that aren't provided by the `ComponentEditor`?

https://github.com/enthought/chaco/blob/fdd858aa6dbc76addb50d011fb81e879ce8e0355/chaco/chaco_plot_editor.py#L63-L67
https://github.com/enthought/chaco/blob/fdd858aa6dbc76addb50d011fb81e879ce8e0355/chaco/chaco_plot_editor.py#L163-L164
https://github.com/enthought/chaco/blob/fdd858aa6dbc76addb50d011fb81e879ce8e0355/chaco/chaco_plot_editor.py#L212-L213

rahulporuri:
closing as we have decided that these can and should be replaced with the enable `ComonentEditor`. Ref #660

PR #648, opened 2021-04-12
Dont hook up listeners in ChacoPlotEditor if the trait is None

rahulporuri:
This PR updates the `ChacoPlotEditor` to not hook up the listeners if the trait is `None` or if the trait doesnt exist on the plot object. This was discovered when running the examples https://github.com/enthought/chaco/blob/fdd858aa6dbc76addb50d011fb81e879ce8e0355/examples/demo/basic/traits_editor.py and https://github.com/enthought/chaco/blob/fdd858aa6dbc76addb50d011fb81e879ce8e0355/examples/demo/advanced/data_stream.py

### DIFF
--- a/chaco/chaco_plot_editor.py
+++ b/chaco/chaco_plot_editor.py
@@ -14,7 +14,7 @@ from enable.api import (
 )
 from enable.trait_defs.ui.api import RGBAColorEditor
 from kiva.trait_defs.kiva_font_trait import KivaFont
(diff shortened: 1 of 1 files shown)


aaronayres35 (approved):
Changes LGTM

I tested and both examples are now working on this branch.

Just wanted to link #640 here though as it is possible we want to remove this module/update or remove these examples along with that

PR #650, opened 2021-04-12
Use #: instead of # when needed

### LINKED ISSUE #615: Use #: instead of # for trait comments.
We would like to convert `#` comments into `#:` so that the trait comments get picked up in the auto-generated API documentation.

aaronayres35:
closes #615 

This PR simply audits the code base for public trait definitions that use # instead of #: and replaces them as needed.  Note some of the modules changed here may soon be deprecated.


rahulporuri (approved):
LGTM

PR #651, opened 2021-04-12
Use @property decorator instead of property function

### LINKED ISSUE #609: Use @property decorator instead of property function
There are a few places in the codebase where we use the old style `property(...)` to define getters and setters for properties - which we can now replace with the `@property` decorator.

https://github.com/enthought/chaco/blob/d7885d5de2f2285414e334ebcdbce20175c87bd4/chaco/shell/plot_window.py#L90-L91

https://github.com/enthought/chaco/blob/d7885d5de2f2285414e334ebcdbce20175c87bd4/chaco/shell/plot_window.py#L210-L211

https://github.com/enthought/chaco/blob/3de7780561fa29e79c887432d3ce408ea82d1614/chaco/tests/test_speedups.py#L58

aaronayres35:
closes #609 

This PR simply updates the uses of the `property` function to use the `@property` decorator.  Also, users of the property now access the property without calling the `get____` or `set___` methods.

### DIFF
--- a/chaco/plugin/workbench_session.py
+++ b/chaco/plugin/workbench_session.py
@@ -37,7 +37,7 @@ class WorkbenchSession(PlotSession):
             window=workbench.active_window,
         )
         new_win.data = self.data
-        new_win.get_container().data = self.data
+        new_win.container.data = self.data
         new_win.session = self
 
         if title is not None:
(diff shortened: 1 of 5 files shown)


rahulporuri (approved):
LGTM

PR #653, opened 2021-04-12
Remove annotated examples from the usermanual

aaronayres35:
This PR simply removes the annotated examples links from the user guide.  They are already linked from the home page of the documentation, and they simply cloud the user manual TOC.


rahulporuri (approved):
LGTM

PR #654, opened 2021-04-12
Use sphinx todo for currently empty sections of docs

aaronayres35:
This PR adds various `.. todo::` statements across the docs to prevent empty sections from being shown in the built docs.

~Note this generates additional sphinx warnings which I am unsure how to resolve currently~
These were warnings of the form "Unexpected Section Title" and commenting out the section titles inside the todo blocks got rid of them


rahulporuri (approved):
LGTM. I don't think we need an issue right now but if we don't address all of the TODO items before the upcoming release (which is possible), we will want to create one.

PR #655, opened 2021-04-12
Remove optparse from example

### LINKED ISSUE #607: Replace optparse with argparse in example
The chaco example https://github.com/enthought/chaco/blob/ef7227706642f7d26042717ad9b60648d7572068/examples/demo/advanced/scalar_image_function_inspector.py makes use of `optparse` which we should replace with `argparse` at the least, if not `click`.

aaronayres35:
closes #607 

This PR removes `optparse` in https://github.com/enthought/chaco/blob/ef7227706642f7d26042717ad9b60648d7572068/examples/demo/advanced/scalar_image_function_inspector.py and also fixes the following error:
```
$ python examples/demo/advanced/scalar_image_function_inspector.py -n 50
2021-04-12 16:29:40.435 Python[68324:22123757] ApplePersistenceIgnoreState: Existing state will not be touched. New state will be written to (null)
Exception occurred in traits notification handler for object: <__main__.PlotUI object at 0x7fee44635200>, trait: num_levels, old value: 15, new value: 50
  File "/Users/aayres/.edm/envs/chaco-test-3.6-pyqt5/lib/python3.6/site-packages/traits/trait_notifiers.py", line 342, in __call__
    self.handler(*args)
  File "examples/demo/advanced/scalar_image_function_inspector.py", line 452, in _num_levels_changed
    self.polyplot.levels = self.num_levels
AttributeError: 'NoneType' object has no attribute 'levels'
```
by adding an `is not None` check to the example.

All of the options for optparse are configurable within the application itself and so it was decided the command line options were unnecessary

### DIFF
--- a/examples/demo/advanced/scalar_image_function_inspector.py
+++ b/examples/demo/advanced/scalar_image_function_inspector.py
@@ -12,11 +12,10 @@ Animation must be disabled (unchecked) before the model can be edited.
 
 
 # Standard library imports
-from optparse import OptionParser
-import sys
 import random
 
 # Major library imports
+import click
 from numpy import array, linspace, meshgrid, nanmin, nanmax, pi, errstate
 
 # Enthought library imports
@@ -450,8 +449,10 @@ class PlotUI(HasTraits):
 
     def _num_levels_changed(self):
         if self.num_levels > 3:
-            self.polyplot.levels = self.num_levels
-            self.lineplot.levels = self.num_levels
+            if self.polyplot is not None:
+                self.polyplot.levels = self.num_levels
+            if self.lineplot is not None:
+                self.lineplot.levels = self.num_levels
 
 
 # HasTraits class that supplies the callable for the timer event.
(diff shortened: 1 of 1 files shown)


rahulporuri (changes requested):
Sorry. I take it back. If we want to contribute this to etsdemo, it will be best if we dont introduce the additional dependency on click. Can you rewrite this to use argparse instead? Sorry for the double work.

aaronayres35:
> Sorry. I take it back. If we want to contribute this to etsdemo, it will be best if we dont introduce the additional dependency on click. Can you rewrite this to use argparse instead? Sorry for the double work.

No worries at all.  I'm happy to make the switch but before I do I wanted to ask: do we event want to support command line arguments at all in that case?  All the options (function, nlevels, and colormap) are configurable via the UI.  I don't exactly see the need of having the command line options, plus it clouds the example with non-chaco code / one is probably looking at the example to learn about chaco not click or aragparse.

However, if we do want the functionality I am happy to just make the switch!

rahulporuri (approved):
LGTM.

Can you add `_default` methods for `function`, `num_levels` and `colormap`? It looks like the changes in `_num_levels_changed` are motivated by the lack of good default values for those traits.

Also, can you update the PR title and description based on the decision we made?

aaronayres35:
> Can you add `_default` methods for `function`, `num_levels` and `colormap`? It looks like the changes in `_num_levels_changed` are motivated by the lack of good default values for those traits.

function already defaults to the default function as its trait definition is `function = Str("tanh(x**2+y)*cos(y)*jn(0,x+y*2)")`.  Likewise for `num_levels = Int(15)`.  `colormap` is a little weird, as `_cmap` defaults to `viridis` as desired and that is what is actually used.  `colormap` defaults to whatever happens to be the first item in `list(default_colormaps.color_map_name_dict.keys())`, but when it changes `_cmap` is updated as needed.
The changes in `_num_levels_changed` have to do with the default for `polyplot` and `lineplot` which get created in `create_plot`.  I think changing this may be a bit more complicated, but I am happy to look into it.  It might be better to do that in a follow up PR though IMO

> Also, can you update the PR title and description based on the decision we made?

Yep, done

rahulporuri:
Still LGTM

Issue #659, opened 2021-04-13
Remove the now unnecessary "use_draw_order" trait?


rahulporuri:
This code has existed in chaco since version 3 so i think it is safe for us to finally remove it.

https://github.com/enthought/chaco/blob/541f3dabbe1a47e6d2c06628d4277910676671fe/chaco/plot_component.py#L65-L77
https://github.com/enthought/chaco/blob/e9af395b5c3bc6364d9c665c9e8d0ddfafc2ac90/chaco/base_plot_container.py#L45-L47

https://github.com/enthought/chaco/blob/e9af395b5c3bc6364d9c665c9e8d0ddfafc2ac90/chaco/axis.py#L235-L236
https://github.com/enthought/chaco/blob/e9af395b5c3bc6364d9c665c9e8d0ddfafc2ac90/chaco/grid.py#L201-L202

PR #661, opened 2021-04-13
Remove unnecessary uses of "use_draw_order"

rahulporuri:
This PR removes unnecessary uses of `use_draw_order` in the code - which is by default `True` - and redefinition of the `use_draw_order` trait on a `PlotComponent` subclass and a commented out piece of code.

Note that I tested the `zoomed_plot` and it still works as expected. You need to do `python -m zoomed_plot.zoom_plot` if you want to run the example (which I think needs to be documented somewhere).

Related to #659


aaronayres35 (approved):
LGTM

PR #665, opened 2021-04-15
Update api module docstring for chaco.api

aaronayres35:
Follow up to #649

This PR populates and move the colormap section to the bottom of the module (although this move may have been unnecessary as I had failed to notice the __all__ previously)
It also moves  `BandedMapper` into the mappers section and `ColormappedSelectionOverlay` into the visual components section

PR #666, opened 2021-04-15
skip test_dont_crash_on_click on windows/pyqt

aaronayres35:
It seems #663 was actually unsuccessful (see #665, #633)

This PR simply skips the test entirely for now to avoid CI headaches on future PRs


rahulporuri (approved):
LGTM

PR #667, opened 2021-04-15
Kill chaco.shell and related code

### LINKED ISSUE #463: Remove chaco.shell
Given that Chaco will never replace Matplotlib, and that `chaco.shell` sees very little use nor is encouraged, we should consider officially deprecating `chaco.shell`.

aaronayres35:
Part of #463 

This PR incrementally removes all uses of `chaco.shell` from the code base.  It may be useful to review commit by commit. 
It starts by removing the entire `chaco.shell` subpackage.  It then removes the `shell` related examples.  It also then removes the `chaco.plugin` sub package as the whole purpose of this subpackage was basically to define an envisage plugin that provides `chaco.shell` functionality.  Without `shell` this plugin becomes useless (AFAICT).  It also removes these files from the list of flake8 ignored files in tox.ini.

This does not completely fix #463 because much of the documentation still uses and refers to `chaco.shell`.  I will open a separate PR for scrapping shell from the docs after this.


rahulporuri (approved):
LGTM

Issue #580, opened 2021-03-22
Remove deprecated modules/classes in 5.0.0 release


rahulporuri:
- [x] [`variable_size_scatterplot`](https://github.com/enthought/chaco/blob/c758a65222a2caef9a3d2c074ae9dd5775a3a77e/chaco/variable_size_scatterplot.py) (#691)
- [x] [`drag_tool`](https://github.com/enthought/chaco/blob/c758a65222a2caef9a3d2c074ae9dd5775a3a77e/chaco/tools/drag_tool.py) (#670)
- [x] [`chaco.chaco_plot_container_editor`](https://github.com/enthought/chaco/blob/c758a65222a2caef9a3d2c074ae9dd5775a3a77e/chaco/chaco_plot_container_editor.py) (#670)
- [x] [`chaco.color_mapper.ColorMapper._map`](https://github.com/enthought/chaco/blob/ad3c77eebee935631f472f1f8ee52d77b5cf71a9/chaco/color_mapper.py#L393-L429)
- [x] [`bg_color` property on `chaco.chaco_plot_editor.ChacoPlotItem`](https://github.com/enthought/chaco/blob/ad3c77eebee935631f472f1f8ee52d77b5cf71a9/chaco/chaco_plot_editor.py#L108-L109) (#688)
- [x] [deprecated traits on `chaco.base_plot_container.BasePlotContainer`](https://github.com/enthought/chaco/blob/ad3c77eebee935631f472f1f8ee52d77b5cf71a9/chaco/base_plot_container.py#L41-L60) (#670)

Plot Frames
-------------
- [x] [`chaco.base_plot_frame`](https://github.com/enthought/chaco/blob/ad3c77eebee935631f472f1f8ee52d77b5cf71a9/chaco/base_plot_frame.py) (#692)
- [x] [`chaco.simple_plot_frame`](https://github.com/enthought/chaco/blob/ad3c77eebee935631f472f1f8ee52d77b5cf71a9/chaco/simple_plot_frame.py) (#692)
- [x] [`chaco.cross_plot_frame`](https://github.com/enthought/chaco/blob/ad3c77eebee935631f472f1f8ee52d77b5cf71a9/chaco/cross_plot_frame.py) (#692)

Zoom Classes
---------------
- [x] [`chaco.tools.simple_zoom`](https://github.com/enthought/chaco/blob/ad3c77eebee935631f472f1f8ee52d77b5cf71a9/chaco/tools/simple_zoom.py) (#670)
[...]

Issue #438, opened 2019-05-15
Clean up the state of the Pan and Zoom Tools


corranwebster:
**Problem Description**

We currently have a plethora of different Pan and Zoom tools (PanTool, PanTool2, ZoomTool, BetterZoomTool, TrackingPanTool, DragZoom, etc.)

We should audit, find out which ones are working as expected, and make them the default (this might be a breaking change, depending on exactly what is needed).

See also #6.

**Expected behavior:**

Developers should have a clear choice for Pan and Zoom tools to use, possibly with one or two clear alternatives that have distinctly different (but common) behaviour.  Tools should be based on Enable tools where practical.

aaronayres35:
As an update:
`BaseZoomTool` and `SimpleZoom` were removed in #670 

Also, the `ZoomTool` currently exposed in `chaco.tools.api` just links to the `BetterSelectingZoom` class also exposed in the `chaco.tools.api`.
`BetterZoom` (also in api) is the base class for `BetterSelectingZoom`. There is also `TrackingZoom` (which subclasses `ZoomTool` which is actually just `BetterSelectingZoom`) in the api. Lastly, there is still `DragZoom` which subclasses `BetterZoom` and `DragTool` from enable.


`pan_tool2.PanTool` is not exposed in the api, and also I'm not seeing it used internally in chaco anywhere currently.

PR #670, opened 2021-04-15
Remove objects/modules which explicitly raised a deprecation warning

aaronayres35:
Part of #580 
Also part of #438 

This PR removes deprecated modules/classes which explicitly raised a deprecation warning.  There are others in which we simply have comments saying things are deprecated.  Perhaps those can be removed entirely, or we should start issuing warnings.  They can be handled in a separate PR.


chaco/tools/tracking_zoom.py
```diff
@@ -24,10 +24,10 @@ def normal_mouse_wheel(self, event):
         if self.enable_wheel and event.mouse_wheel != 0:
             if event.mouse_wheel > 0:
                 # zoom in
-                zoom = 1.0 / (1.0 + 0.5 * self.wheel_zoom_step)
+                zoom = 2.0 / (1.0 + self.zoom_factor)
             elif event.mouse_wheel < 0:
                 # zoom out
-                zoom = 1.0 + 0.5 * self.wheel_zoom_step
+                zoom = (1.0 + self.zoom_factor) / 2.0
```
aaronayres35, line 30: Note that `self.wheel_zoom_step = self.zoom_factor - 1.0`

chaco/tools/simple_zoom.py
```diff
@@ -1,649 +0,0 @@
-""" Defines the SimpleZoom class.
```
rahulporuri, line 1: looks like this too was moved into `enable` as the viewport zoom tool and modified.

When we do the release, i think it will be good to highlight the fact that there are classes in enable which can be used by end users if they can't migrate to the chaco ZoomTool immediately.

chaco/tools/simple_zoom.py
```diff
@@ -1,649 +0,0 @@
...
-from enable.api import ColorTrait, KeySpec
-from traits.api import Bool, Enum, Float, Instance, Int, Str, Trait, Tuple
-
-# Chaco imports
-from chaco.abstract_overlay import AbstractOverlay
-from .base_zoom_tool import BaseZoomTool
-from .tool_history_mixin import ToolHistoryMixin
-
-
-class SimpleZoom(AbstractOverlay, ToolHistoryMixin, BaseZoomTool):
```
rahulporuri, line 23: looks like we're using the `SimpleZoom` class in one of the examples - see chaco tutorial, "dynamically controlling interactions" section where `SimpleZoom` is one of the options in the check list editor

examples/demo/canvas/canvas.py
```diff
@@ -228,15 +228,13 @@ def clone_plot(clonetool, drop_position):
         "axis",
         "enable_wheel",
         "drag_button",
-        "wheel_zoom_step",
```
rahulporuri, line 231: i thought the plot canvas examples were broken anyway so i dont care much about the changes made to this example.

chaco/tools/better_selecting_zoom.py
```diff
@@ -38,14 +37,6 @@ class BetterSelectingZoom(AbstractOverlay, BetterZoom):
...
-    # deprecated interaction controls, used for API compatability with
-    # SimpleZoom
-    # -------------------------------------------------------------------------
-
-    #: Conversion ratio from wheel steps to zoom factors.
-    wheel_zoom_step = Property(Float, observe="zoom_factor")
-
     #: The key press to enter zoom mode, if **always_on** is False.  Has no effect
     #: if **always_on** is True.
     enter_zoom_key = Instance(KeySpec, args=("z",))
```
rahulporuri, line 42: Aren't `enter_zoom_key` and `exit_zoom_key` also part of the deprecated interface? I see those traits on the `SimpleZoom` class. Sadly, they weren't explicitly marked as deprecated.

But then again, these traits do feel like they belong to this class/api.

chaco/tools/tracking_zoom.py
rahulporuri: we don't need the `2.0`, `1.0` anymore on python 3

rahulporuri (approved):
LGTM with a couple of comments/questions

chaco/tools/simple_zoom.py
aaronayres35: Ah looks like `ZoomTool` is imported from `chaco.tools.api` but `"SimpleZoom"` is what was listed for the checklist editor.  I've updated the item in the checklist to be `"ZoomTool"`

chaco/tools/better_selecting_zoom.py
aaronayres35: Yeah these weren't throwing warnings, and they also appear to be being explicitly used by the class so I figured they would need to stay

Issue #232, opened 2014-10-29
Colormapped scatterplot with custom marker render fails


corranwebster:
See tests added in #231 

At least some of the problem is at the Enable level with improper calls to `scale_ctm`, but there is something further wrong even if that is fixed.

PR #672, opened 2021-04-15
Use draw_marker_at_points if possible otherwise use draw_path_at_points in ColormappedScatterPlot

### LINKED ISSUE #425: new markers (star, hexagon...) available in ScatterPlot but not in ColormappedScatterPlot
Hi there,

New ScatterPlot markers as star, hexagon, hexagon2, pentagon and so on are not available with ColormappedScatterPlot.

For instance, if you run Chaco demo example scatter.py in basic/, it works well if you change line
"marker='circle'" by "marker='star'".
But if you run basic/cmap_scatter.py with the same modification, it displays... nothing.

So my question is : could it be possible to implement these new markers in ColormappedScatterPlot please? 

Thanks in advance,

Cheers,

Debian Linux 64bits unstable, Enthought code from git, Python 2.7, wxPython 3.0.2, Chaco 4.7.1.dev45, Traits 4.7.0.dev94 & TraitsUI 6.1.0.dev22

aaronayres35:
fixes #425 
This can close issue #232 as well after https://github.com/enthought/enable/pull/782 is merged

 In enable:
https://github.com/enthought/enable/blob/master/enable/markers.py
some markers have a defined `kiva_marker` while others simply define an `_add_to_path` method detailing how to draw the marker manually. If the marker has `kiva_marker` we can use `draw_marker_at_points`, but if it doesn't we need to fall back to manually drawing the path via the `_add_to_path` method and calling `draw_path_at_points`.

### DIFF
--- a/chaco/colormapped_scatterplot.py
+++ b/chaco/colormapped_scatterplot.py
@@ -398,8 +398,13 @@ class ColormappedScatterPlot(ScatterPlot):
                     gc, "draw_marker_at_points"
                 ) and self.marker not in (
                     "custom",
-                    "circle",
-                    "diamond",
+                    "left_triangle",
+                    "right_triangle",
+                    "pentagon",
+                    "hexagon",
+                    "hexagon2",
+                    "star",
+                    "cross_plus"
                 ):
                     draw_func = lambda x, y, size: gc.draw_marker_at_points(
                         [[x, y]], size, marker_cls.kiva_marker
--- a/chaco/colormapped_scatterplot.py
+++ b/chaco/colormapped_scatterplot.py
@@ -17,7 +17,7 @@ from numpy import (
 )
 
 # Enthought library imports
-from kiva.constants import STROKE
+from kiva.api import NO_MARKER, STROKE
 from traits.api import Dict, Enum, Float, Instance, observe
 from traitsui.api import Item, RangeEditor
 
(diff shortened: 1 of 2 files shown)


chaco/tests/test_colormapped_scatterplot.py
```diff
@@ -86,3 +86,11 @@ def test_colormap_updated(self):
...
         self.color_mapper.updated = True
         self.assertFalse(self.scatterplot.draw_valid)
+
+    # regression test for enthought/chaco#425
+    def test_non_kiva_marker(self):
+        self.scatterplot.marker = "star"
+
+        self.gc.render_component(self.scatterplot)
+        actual = self.gc.bmp_array[:, :, :]
+        self.assertFalse(alltrue(actual == 255))
```
aaronayres35, line 96: I was being dumb before, it was trivial to write this regression test...

aaronayres35: Note this PR also fixes the test `test_scatter_custom` above once https://github.com/enthought/enable/pull/782 is pushed through that test can be unskipped.

chaco/colormapped_scatterplot.py
```diff
@@ -393,14 +390,10 @@ def _render_bruteforce(self, gc, points):
                 marker_size = marker_size[self._cached_point_mask]
             mode = marker_cls.draw_mode
 
-            if marker_cls != "custom":
+            if self.marker != "custom":
```
aaronayres35, line 393: marker_cls is not a string, but the actual class e.g. `<class 'enable.markers.CustomMarker'>`

chaco/colormapped_scatterplot.py
```diff
@@ -434,7 +427,7 @@ def draw_func(x, y, size):
                     draw_func(x[i], y[i], size)
 
             else:
-                path = marker_cls.custom_symbol
+                path = self.custom_symbol
```
aaronayres35, line 430: `custom_symbol` is a trait on the `ColormappedScatterPlot`, not the marker. For example use, see test `test_scatter_custom`.

aaronayres35, line 430: `custom_symbol` is a trait on the `ColormappedScatterPlot`, not the marker. For example use, see test `test_scatter_custom`.

rahulporuri (approved):
LGTM. Do you want to open a separate issue about removing the test skip once we have a new enable?

aaronayres35:
I've opened #686

PR #673, opened 2021-04-16
Use stacked layout utils from enable

rahulporuri:
This PR updates the stacked plot containers in chaco to use utils from enable. The utils seem to have been moved to enable a long time back but we just werent using them in chaco.

### DIFF
--- a/chaco/plot_containers.py
+++ b/chaco/plot_containers.py
@@ -32,6 +32,7 @@ from enable.simple_layout import (
     simple_container_get_preferred_size,
     simple_container_do_layout,
 )
+from enable.stacked_layout import stack_layout, stacked_preferred_size
 
 try:
     from enable.api import ConstraintsContainer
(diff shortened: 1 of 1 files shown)


chaco/plot_containers.py
```diff
@@ -136,145 +137,12 @@ def get_preferred_size(self, components=None):
...
-                max_other_size + self.vpadding,
-            )
-        else:
-            self._cached_preferred_size = (
-                max_other_size + self.hpadding,
-                total_size + self.vpadding,
-            )
-
-        return self._cached_preferred_size
+        return stacked_preferred_size(self, components=components)
```
rahulporuri, line 140: Note that there is some duplication here.

`stacked_preferred_size` does not handle the case where `resizable=""` - and we might want to update the underlying enable util to handle this case - but until then, we will have to explicitly handle the `if self.fixed_preferred_size is not None:` case AND the `if self.resizable == "":` case - and implicitly handle the `if self.fixed_preferred_size is not None:` case again inside the enable util, which will not do anything. under the circumstances.

PR #674, opened 2021-04-16
[WIP] Refactor chaco plot containers

rahulporuri:


chaco/plot_containers.py
```diff
@@ -19,27 +19,25 @@
...
+from enable.stacked_layout import stack_layout, stacked_preferred_size
 
 try:
     from enable.api import ConstraintsContainer
 except ImportError:
     ConstraintsContainer = None
 
 # Local relative imports
-from .base_plot_container import BasePlotContainer
+from .plot_component import DEFAULT_DRAWING_ORDER
```
rahulporuri, line 40: we might want to move `DEFAULT_DRAWING_ORDER` somewhere else/somewhere better e.g. `chaco.base` instead of leaving it in `chaco.plot_component`

chaco/plot_containers.py
```diff
@@ -136,176 +124,7 @@ def get_preferred_size(self, components=None):
...
-    stack_order = Enum("left_to_right", "right_to_left")
-
-    #: The amount of space to put between components.
-    spacing = Float(0.0)
-
-    #: The vertical alignment of objects that don't span the full height.
-    valign = Enum("bottom", "top", "center")
-
-    _cached_preferred_size = Tuple(transient=True)
+        return stacked_preferred_size(self, components=components)
```
rahulporuri, line 127:
```
        if self.resizable == "":
            self._cached_preferred_size = self.outer_bounds[:]
            return self.outer_bounds
```

isn't handled by enables `stacked_preferred_size`. If we can update the function in enable, after the next release of enable, we will be able to get rid of this method completely. both in this class and in `VPlotContainer`

chaco/plot_containers.py
```diff
@@ -321,51 +140,59 @@ def _do_layout(self):
         else:
             align = "max"
 
-        return self._do_stack_layout(components, align)
+        return stack_layout(self, components=components, align=align)
```
rahulporuri, line 143: there are bugs in `HStackedContainer` and the `HStackedContainer._do_layout` - if those get fixed, we will be able to remove this `HPlotContainer._do_layout` method and `HPlotContainer.valign` trait.

chaco/plot_containers.py
```diff
@@ -321,51 +140,59 @@ def _do_layout(self):
...
+
+    ### Persistence ###########################################################
 
-        return self._do_stack_layout(components, align)
+    # PICKLE FIXME: blocked with _pickles, but not sure that was correct.
+    def __getstate__(self):
+        state = super(VStackedContainer, self).__getstate__()
+        if "stack_index" in state:
+            del state["stack_index"]
+        return state
```
rahulporuri, line 192: If we can push this `__getstate__` down into enable, we can get rid of these in the chaco subclasses. Note that this PR also removes a bunch of traits in the subclasses which were marked as `transient` in the subclasses but they are not `transient` in the superclasses e.g. `_cached_preferred_size` (which isn't even defined on the baseclasses in the first place)

chaco/plot_containers.py
```diff
@@ -321,51 +140,59 @@ def _do_layout(self):
...
 
 
-class GridPlotContainer(BasePlotContainer):
```
rahulporuri, line 368: Note that we are now in a place where we can get rid of `BasePlotContainer` completely.

chaco/plot_containers.py
```diff
@@ -19,27 +19,25 @@
...
     Trait,
     Tuple,
     Int,
 )
-from enable.simple_layout import (
-    simple_container_get_preferred_size,
-    simple_container_do_layout,
-)
+from enable.api import Container, OverlayContainer
+from enable.stacked_container import HStackedContainer, VStackedContainer
```
rahulporuri, line 31: `enable` needs to expose these via the api module.

aaronayres35: ref: https://github.com/enthought/enable/pull/779

chaco/plot_containers.py
aaronayres35: ref: https://github.com/enthought/enable/pull/778

chaco/plot_containers.py
aaronayres35: https://github.com/enthought/enable/pull/780

PR #675, opened 2021-04-16
Import color palette from chaco.api directly

rahulporuri:
Some of the examples import `cbrewer` from `chaco.api` as `COLOR_PALETTE` from `chaco.example_support`. I am not sure `chaco.example_support` should be a public module and I don't think we should be promoting this usage. This PR replaces all instances of `from chaco.example_support import COLOR_PALETTE` with `from chaco.api import cbrewer as COLOR_PALETTE`.


aaronayres35 (approved):
LGTM

Eventually when we flake8 clean everything we will want imports in alphabetical order but that can be dealt with in a later PR.

PR #676, opened 2021-04-16
Fix a couple of demos

rahulporuri:
- `multi_line_plot.py`
- `multi_line_plot_demo.py`
- `javascript_hover_tools.py`

The first two were failing earlier because of a bad listener hookup. The latter was failing because of the use of non-existent methods.

### DIFF
--- a/chaco/multi_line_plot.py
+++ b/chaco/multi_line_plot.py
@@ -143,7 +143,7 @@ class MultiLinePlot(BaseXYPlot):
         observe=[
             "global_min",
             "global_max",
-            "data",
+            "value._data",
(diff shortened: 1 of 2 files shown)


aaronayres35 (approved):
LGTM
All the demos launch without errors and work as expected now.
I ran the `javascript_hover_tools.py ` and my browser did not immediately launch, but I was able to open the html file manually and things seemed to be working

rahulporuri:
> ... and my browser did not immediately launch, but I was able to open the html file manually and things seemed to be working

Hmm. Weird. The browser opened up for me as soon the script finished running.

PR #677, opened 2021-04-16
Remove "chaco.scatter_markers" module

rahulporuri:
This PR removes the `chaco.scatter_markers` module and updates one import in chaco to use `enable.api` instead. The file is also removed from the list of ignored files for flake8.

As can be seen from the module docstring, the contents of this module were moved to enable so the imports should also be updated to use enable.


aaronayres35 (approved):
LGTM

aaronayres35:
ref #580

Issue #660, opened 2021-04-13
Replace use of ChacoPlotItem with just Item and ComponentEditor


rahulporuri:
There are a few examples which use `ChacoPlotItem` instead of the enable `ComponentEditor`. It is possible that the `ComponentEditor` came after the chaco `ChacoPlotItem` and related objects. I think we can rewrite the examples to use `ComponentEditor` instead of the chaco classes.

See related #640

rahulporuri:
FTR, I spent a little bit of time on this and this is not a trivial task.

PR #680, opened 2021-04-16
Update van der waal example

aaronayres35:
part of #660 

This PR updates the Van Der Waal example in the documentation.  It removes the use of `ChacoPlotItem` and the like with `ComponentEditor`, replaces use of `on_trait_change` with `observe`, removes development setup section, cleans up wording, and adds the code for the example in `examples/demo/vanderwaals.py`.


docs/source/user_manual/tutorial_van_der_waal.rst
```diff
@@ -285,24 +320,26 @@ Source Code
 The final version on the program, `vanderwaals.py` ::
```
rahulporuri, line 320: Let's not explicitly copy the code from the python file here - instead just use `literalinclude`. Also, it would be awesome if we provide a way to click on `vanderwaals.py` and see the example on github. I think that's possible - not sure how.

docs/source/user_manual/tutorial_van_der_waal.rst
```diff
@@ -167,26 +174,48 @@ Next, there are two programing tasks to complete:
...
+define a separate listener as follows::
+
+    @observe("plot_type")
+    def _update_plot_type(self, event):
+        old_plot_type, new_plot_type = event.old, event.new
+
+        self.plot.delplot(old_plot_type)
+        self.plot.plot(
+            ("x", "y"), type=new_plot_type, name=new_plot_type, color="blue"
+        )
```
rahulporuri, line 207: so when i change `plot_type`, i need to change some other variable before the plot is updated i.e. if i change from line to scatter, i need to change temperature before the plot changes from line plot to scatter plot.

examples/demo/vanderwaals.py
```diff
@@ -0,0 +1,101 @@
...
+    Float,
+    HasTraits,
+    Instance,
+    observe,
+    Property,
+    Range,
+)
+from traitsui.api import Item, UItem, View
+from chaco.api import ArrayPlotData, Plot
+from enable.api import ComponentEditor
```
rahulporuri, line 17: nit-pick : for users of ETS, i think all of ETS comes under the same category so `traits` and `traitsui` come after `chaco` and `enable`. Also, numpy needs to come before `traits`.

nit-nit-pick : i don't understand why we do `from numpy import arange` where as the community normally does `import numpy as np; np.arange`.

rahulporuri (approved):
Almost LGTM - there is one minor bug and a small suggestion.

docs/source/user_manual/tutorial_van_der_waal.rst
aaronayres35: Ah I wasn't observing this locally.  Thanks for pointing this out. I tried adding `self.plot.invalidate_and_redaw()`. Lmk if this fixes it for you

docs/source/user_manual/tutorial_van_der_waal.rst
aaronayres35: I've tried doing both of these things. The link to github is tricky because the file doesn't exist yet, it is bing introduced in this PR.  The link I used is where the file _will_ be located, but currently it is a broken link if you try to test it on this PR... 
I am not sure the best way to go about this

docs/source/user_manual/tutorial_van_der_waal.rst
aaronayres35: FWIW I was seeing something similar to what you describe while working on the hyetograph example and `invalidate_and_redraw` fixed it

rahulporuri:
still LGTM

PR #684, opened 2021-04-19
Remove deprecated "ColorMapper._map" method

rahulporuri:
part of #580

This PR removes the deprecated `ColorMapper._map` method.


aaronayres35 (approved):
LGTM

It appears we haven't been raising a deprecation warning for use of the method, but its internal use has been removed and it is a privately named method so it should be fine to remove

PR #685, opened 2021-04-19
Delete sample_template.plt

rahulporuri:
Not sure when the code that used this code got removed from the codebase but this file isnt used anywhere in the 3.0.0 tag. Part of #580.


aaronayres35 (approved):
LGTM

PR #691, opened 2021-04-19
Officially deprecated VariableSizeScatterPlot

aaronayres35:
Part of #580 

`VariableSizeScatterPlot` has been commented as being deprecated, but no warning was raised.  This PR officially deprecates it by adding an explicit warning when importing it.

It also adds a trivial test of the deprecation warning which may or may not be overkill...


rahulporuri:
Same here. I don't think there's a point to deprecating it officially now and removing it in a later release. I don't see any meaningful users of `VariableSizeScatterPlot` on GitHub.

examples/demo/advanced/cmap_variable_sized_scatter.py
```diff
@@ -1,6 +1,5 @@
 """
 Draws a scatterplot of a set of random points of variable size and color.
- - This uses the non-standard renderer, VariableSizeScatterPlot
```
rahulporuri, line 3: looks like we should rename this file - it doesn't use `VariableSizeScatterPlot` anymore - just the `ColormappedScatterPlot`

rahulporuri (approved):
LGTM with one minor suggestion

examples/demo/advanced/cmap_variable_sized_scatter.py
aaronayres35: It is still an example of a scatter plot that uses variable sized markers though.  It just uses this functionality on the `ScatterMap` class directly

I'm not sure what we would rename it to

rahulporuri: fair enough. the name conveys behavior, not which plot type it is. never mind. let's leave it as is.

PR #692, opened 2021-04-19
Remove PlotFrames

aaronayres35:
Part of #580 

This PR adds removes the various plot frame classes from the code base.

~Unfortunately, since these objects are exposed in `chaco.api` now just importing from chaco.api will raise a deprecation warning...Perhaps these warnings should be raised when instantiating a plot frame object?~

~Edit: I moved the warnings to occur inside the `__init__` methods and updated the tests.~


rahulporuri (changes requested):
I would just remove them outright. I don't think it's worth deprecating them now and waiting for another rleease to remove them. I'll check if there are any users of the plot frame classes.

rahulporuri:
I think it's safe to remove them outright. I only found one meaningful/new-ish use of `CrossPlotFrame` internally when I search on GitHub.

corranwebster:
They can be removed outright I think - they have been superseded by the various plot layout classes.

docs/chaco_layout.txt
```diff
@@ -3,25 +3,8 @@ Notes on layout
...
-
-Containers can be placed within these slots, and other containers can be nested
-within those, so that within each slot, a very flexible layout can be achieved.
-For example, geophysical "log plots" generally consist of a SimplePlotFrame with
-a horizontal plot container (HPlotContainer) in the "center" slot.  Tracks in
-the horizontal container are VPlotContainers, with GeoAxis components stacked
-above a single plot area, which is an OverlayPlotContainer.  (The GeoTrack class
-is actually a subclass of VPlotContainer that automatically create GeoAxis objects
-as GeoPlots are added to the plot area.)
-
```
aaronayres35, line 24: There are other .txt files like this in the docs directory but they aren't actually a part of the documentation

rahulporuri (approved):
LGTM

PR #693, opened 2021-04-20
Rename test so it is discovered by unittest discover

rahulporuri:
This PR renames a test module so that it is discovered by `unittest discover`. Not sure how we missed it earlier when we renamed the rest of the testsuite.


aaronayres35 (approved):
LGTM

I compared appveyor CI log on this PR and others and number of tests run went from 547 to 568

PR #694, opened 2021-04-20
[Do not merge] Line length fixes to docs

rahulporuri:
This PR makes line length fixes to a couple of documentation files.

### DIFF
--- a/docs/chaco_layout.txt
+++ b/docs/chaco_layout.txt
@@ -3,15 +3,7 @@ Notes on layout
 
 The process of layout involves setting the size and position of components.
 All plot components must define two methods: layout() and get_preferred_size().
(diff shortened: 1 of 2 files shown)


aaronayres35 (commented):
These txt files are not actually in the documention and on Github it looks they havent been touched for 10+ years.  Perhaps we can delete them entirely? (After going through them looking for anything we might want to transfer over to the real docs

corranwebster:
These look like old design docs.  They may be of some historical interest to know where people were thinking of taking things, but they are misleading if they can be confused with the state of things now.

It's probably safe to remove them, or if we are feeling a bit more cautious, move them into a "historical design" sub-folder.

PR #695, opened 2021-04-20
Cleanup ci module and setup

rahulporuri:
This PR makes a number of changes to the ci module and setup script

- removes use of `use_2to3` in setup script
- fixes flake8 errors and failures in setup script
- fixes flake8 errors in ci module
- moves `coverage` from requirements file to edmtool so it can be installed via edm instead of pypi
- removes unused `travis-ci-requirements.txt`
- update tox config file to remove fixed up files from flake8 ignore list

### DIFF
--- a/travis-ci-requirements.txt
+++ b/travis-ci-requirements.txt
@@ -1,6 +1,0 @@
-numpy
-coverage
-pillow
-git+http://github.com/nucleic/kiwi.git#egg=kiwi
-traits
-traitsui
--- a/ci/edmtool.py
+++ b/ci/edmtool.py
@@ -75,6 +75,7 @@ supported_combinations = {
 }
 
 dependencies = {
+    "coverage",
     "mock",
     "numpy",
     "pandas",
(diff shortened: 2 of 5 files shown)


travis-ci-requirements.txt
```diff
@@ -1,6 +0,0 @@
-numpy
-coverage
-pillow
-git+http://github.com/nucleic/kiwi.git#egg=kiwi
```
aaronayres35, line 4: This file looks like it was unused, but I don't see these anywhere else explicitly. Im assuming they come from enable, and so don't need to be listed as deps

aaronayres35 (approved):
LGTM

travis-ci-requirements.txt
rahulporuri: not sure where these came from. i didn't bother looking for when this was added to the repo. but you're probably right - they were possible needed for source installs of enable, which are handled by the ci module now.

PR #697, opened 2021-04-20
Delete "test_style.pss"

rahulporuri:
This PR removes the unused `test_style.pss` file. I am not sure who ever used it or when we stopped needing it.


aaronayres35 (approved):
LGTM

PR #703, opened 2021-04-20
Update "super" usage

rahulporuri:
On Python 3, we dont need to pass any arguments to `super` in the usual case where we are simply calling `super` with the defining class. This PR removes such instances.

Note that the changes were made using automated regex-based search and replace, similar to https://github.com/enthought/traits/pull/1280. After the changes were made, we checked each of the changes to ensure that we not passing a different class when calling `super` than the defining class e.g. calling `super(A, self)` in class `B` where `B` inherits from `A`.  There were a few such instances discovered in the code - which will be addressed in a later PR.

### DIFF
--- a/chaco/abstract_controller.py
+++ b/chaco/abstract_controller.py
@@ -15,7 +15,7 @@ class AbstractController(Interactor):
 
     def __init__(self, component, *args, **kw):
         self.component = component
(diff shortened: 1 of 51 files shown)


aaronayres35 (approved):
LGTM

Issue #658, opened 2021-04-13
Use a plot default method instead of overriding the init method in examples


rahulporuri:
The chaco examples create the displayed plot object in atleast two different ways - they either override the `__init__` method of the underlying traits class to create and store the plot object or they define a `plot` trait and a `_plot_default` method which creates the plot object.

I think the latter way is preferred over the former - we don't want to unnecessarily override the `__init__` method.

https://github.com/enthought/chaco/blob/e5a85948e41720cc3ba2ad0e03a1e375f657f32b/examples/user_guide/h_plot_container.py#L20-L36

https://github.com/enthought/chaco/blob/e5a85948e41720cc3ba2ad0e03a1e375f657f32b/examples/user_guide/power_function_example.py#L26-L48

PR #704, opened 2021-04-20
Use a plot default method instead of overriding the init method in examples

aaronayres35:
Part of #658 

This PR replaces the `__init__` method in many examples with the `_plot_default` approach.  This is not all the examples, but the remaining can be done in a follow up PR.


rahulporuri (approved):
LGTM

PR #705, opened 2021-04-21
Replace custom "assert_raises" with "unittest.TestCase.assertRaises"

rahulporuri:
This PR replaces the use of the custom `assert_raises` context manager with the `unittest.TestCase.assertRaises` and removes the custom `assert_raises` context manager.


aaronayres35 (approved):
LGTM

PR #706, opened 2021-04-21
Fix potentially buggy calls to "super"

rahulporuri:


chaco/array_plot_data.py
```diff
@@ -59,7 +59,7 @@ def __init__(self, *data, **kw):
         conventions, but was it chosen for convenience, since the raison d'etre
         of this class is convenience.
         """
-        super(AbstractPlotData, self).__init__()
+        super(ArrayPlotData, self).__init__()
```
corranwebster, line 62: Since we are only supporting Python 3+ now, we can use `super()` and avoid these problems entirely.

rahulporuri: that is the final plan - i wanted to make smaller changes incrementally to ensure that the behavior is still the same.

rahulporuri, line 62:
```suggestion
        super().__init__()
```

chaco/base_data_range.py
```diff
@@ -17,7 +17,7 @@ class BaseDataRange(AbstractDataRange):
     # ------------------------------------------------------------------------
 
     def __init__(self, *datasources, **kwtraits):
-        super(AbstractDataRange, self).__init__(**kwtraits)
+        super(BaseDataRange, self).__init__(**kwtraits)
```
rahulporuri, line 20:
```suggestion
        super().__init__(**kwtraits)
```

chaco/tools/tests/test_image_inspector.py
```diff
@@ -83,7 +83,7 @@ class TestImageInspectorToolGray(BaseImageInspectorTool, TestCase):
     def setUp(self):
         values = np.arange(4).reshape(2, 2)
         self.plot = create_image_plot(values)
-        super(TestImageInspectorToolGray, self).setUp()
+        BaseImageInspectorTool.setUp(self)
```
aaronayres35, line 86: Why exactly can't this be `super().setUp()`?  I notice the tests all still pass if that change is made.

aaronayres35 (approved):
LGTM just one question

chaco/tools/tests/test_image_inspector.py
rahulporuri: we discussed this in person - this is part convention and part best practice afaik - especially because we have a few places where we need to call `setUp` on more than one baseclass.

PR #709, opened 2021-04-21
Use a plot default method instead of overriding the init method in examples part 2

### LINKED ISSUE #658: Use a plot default method instead of overriding the init method in examples

aaronayres35:
closes #658 

There are still some examples which define an `__init__` method.  However they either are ding something where they likely need to keep it or they belong to `examples/tutorials/scipy2008/` in which I didn't touch them.

This PR also fixes many of these examples which previously had been broken (trivial changes eg defining `_create_component` not `_create_window`, or using ints instead of floats).

### DIFF
--- a/examples/demo/advanced/spec_waterfall.py
+++ b/examples/demo/advanced/spec_waterfall.py
@@ -231,9 +231,8 @@ class Demo(HasTraits):
         handler=DemoHandler,
     )
 
-    def __init__(self, **traits):
-        super(Demo, self).__init__(**traits)
-        self.plot = _create_plot_component(self.controller)
+    def _plot_default(self):
+        return _create_plot_component(self.controller)
 
     def edit_traits(self, *args, **kws):
         # Start up the timer! We should do this only when the demo actually
--- a/examples/demo/advanced/spec_waterfall.py
+++ b/examples/demo/advanced/spec_waterfall.py
@@ -102,7 +102,7 @@ def _create_plot_component(obj):
     # Setup the spectrum plot
     frequencies = linspace(0.0, float(SAMPLING_RATE) / 2, num=NUM_SAMPLES / 2)
     obj.spectrum_data = ArrayPlotData(frequency=frequencies)
-    empty_amplitude = zeros(NUM_SAMPLES / 2)
+    empty_amplitude = zeros(NUM_SAMPLES // 2)
     obj.spectrum_data.set_data("amplitude", empty_amplitude)
 
     obj.spectrum_plot = Plot(obj.spectrum_data)
@@ -135,7 +135,7 @@ def _create_plot_component(obj):
     time_range.high = 0.2
 
     # Spectrogram plot
-    values = [zeros(NUM_SAMPLES / 2) for i in range(SPECTROGRAM_LENGTH)]
+    values = [zeros(NUM_SAMPLES // 2) for i in range(SPECTROGRAM_LENGTH)]
     p = WaterfallRenderer(
         index=spec_renderer.index,
         values=values,
@@ -178,7 +178,7 @@ def get_audio_data():
     audio_data = fromstring(stream.read(NUM_SAMPLES), dtype=short)
     stream.close()
     normalized_data = audio_data / 32768.0
-    return (abs(fft(normalized_data))[: NUM_SAMPLES / 2], normalized_data)
+    return (abs(fft(normalized_data))[: NUM_SAMPLES // 2], normalized_data)
 
 
 # HasTraits class that supplies the callable for the timer event.
--- a/examples/demo/basic/image_from_file.py
+++ b/examples/demo/basic/image_from_file.py
@@ -87,24 +87,6 @@ class DemoView(HasTraits):
     # Public 'DemoView' interface
     # ---------------------------------------------------------------------------
 
-    def __init__(self, *args, **kwargs):
-        super(DemoView, self).__init__(*args, **kwargs)
-
-        # Create the plot object, set some options, and add some tools
-        plot = self.plot = Plot(self.pd, default_origin="top left")
-        plot.x_axis.orientation = "top"
-        plot.padding = 50
-        plot.padding_top = 75
-        plot.tools.append(PanTool(plot))
-        zoom = ZoomTool(component=plot, tool_mode="box", always_on=False)
-        plot.overlays.append(zoom)
-
-        # Load the default image
-        self._load()
-
-        # Plot the image plot with this image
-        self.plot.img_plot("imagedata")
-
     def default_traits_view(self):
         """Returns the default view to use for this class."""
         # NOTE: I moved the view to this method so we can declare a handler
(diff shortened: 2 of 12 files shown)


examples/demo/financial/correlations.py
```diff
@@ -197,7 +195,7 @@ def _selections_updated(self, event):
             )
             self.corr_plot.request_redraw()
 
-    @observe("sym1,sym2")
+    @observe("symbols,sym1,sym2")
```
aaronayres35, line 198: this was needed to get the plot to show up immediately when the application launches.  Otherwise a plot would only appear after changing sym1 or sym2.

examples/demo/canvas/cliptest.py
```diff
@@ -70,7 +67,7 @@ def moving_mouse_leave(self, event):
 
 
 class MainFrame(DemoFrame):
-    def _create_window(self):
+    def _create_component(self):
```
rahulporuri, line 70: `canvas.py` in the canvas examples directory also needs the same fix - after which i am running into a very interesting failure mode.

now that atleast one of the canvas examples is fixed, should we close #576 ?

rahulporuri: > canvas.py in the canvas examples directory also needs the same fix - after which i am running into a very interesting failure mode.

it looks like you got the canvas example running in #254 . it'll be good to fix it up in the master branch as wekk.

examples/demo/updating_plot/updating_plot5.py
```diff
@@ -32,27 +32,32 @@
...
+    def _plot_default(self):
+        if type(self.x_values) == ArrayDataSource:
+            self.x_values = self.x_values.get_data()[:]
             plot = create_line_plot(
-                (x, self.y_values),
-                color=color,
-                bgcolor=bgcolor,
+                (self.x_values, self.y_values),
+                color=self.color,
+                bgcolor="none",
```
rahulporuri, line 48: setting bgcolor to `none` here unlike the other examples where it is set to `white` makes the plot hard to read. on windows, the background was grey/black when it is set to `none`

examples/demo/updating_plot/updating_plot2.py
```diff
@@ -96,7 +103,7 @@ def _create_window(self):
 
         self.timer = Timer(100.0, self.onTimer)
         self.container = container
-        return Window(self, -1, component=container)
+        return container
 
     def onTimer(self, *args):
         for plot in self.animated_plots:
```
rahulporuri, line 109: this example (and other `updating_plot*.py` examples) work as expected - but the window sizes are too small. From the looks of it, the `size` argument passed to `demo_main` isn't used so simply asking for a larger size there does not help.

rahulporuri (approved):
Mostly LGTM

examples/demo/updating_plot/updating_plot2.py
```diff
@@ -69,13 +74,13 @@ def timer_tick(self):
         self.current_index += self.increment
         if self.current_index > self.numpoints:
             self.current_index = self.numpoints
-        self.plot.index.set_data(self.x_values[: self.current_index])
-        self.plot.value.set_data(self.y_values[: self.current_index])
+        self.plot.index.set_data(self.x_values[: int(self.current_index)])
```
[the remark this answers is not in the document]
aaronayres35: ah you are right, I actually made this as the fix first when I first encountered the issue, but then realized it could be fixed at the source.  I can undo this change

examples/demo/updating_plot/updating_plot2.py
```diff
@@ -36,15 +36,19 @@
 
 
 class AnimatedPlot(HasTraits):
-    def __init__(self, x, y, color="blue", bgcolor="white"):
-        self.x_values = x[:]
-        self.y_values = y[:]
+    x_values = Array()
+    y_values = Array()
+    color = Str()
```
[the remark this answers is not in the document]
aaronayres35: This is because I noticed it was never being set any time the class was instantiated.  ie we would do something like `AnimatedPlot(x, jn(i, x), color)` and `bgcolor` would always just be he default.
If it is not used in this example code, and is not configurable in the UI I figured it could just be fixed

examples/demo/updating_plot/updating_plot3.py
```diff
@@ -32,23 +32,28 @@
...
-                color=color,
-                bgcolor=bgcolor,
+                (self.x_values, self.y_values),
+                color=self.color,
+                bgcolor="white",
                 add_grid=True,
                 add_axis=True,
             )
         else:
-            self.x_values = x[:]
```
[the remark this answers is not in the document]
aaronayres35: I was confused as well, these are also np arrays so shouldn't `x[:]` just be a ref to `x`?

Ie if `x` is a np array, and  set `y= x[:]` and change `y` `x` changes as well.

examples/demo/updating_plot/updating_plot2.py
aaronayres35: yeah not sure why that exists anymore, it can probably be removed in enable. 

But also, Should we even be importing from `enable.example_support` in the first place

rahulporuri: > But also, Should we even be importing from enable.example_support in the first place

I would say no. It's a small wrapper (`demo_main`/`DemoFrame`) that adds an unnecessary abstraction in the middle IMO, especially in a user-facing example. But let's discuss and handle that separately.

examples/demo/advanced/spec_waterfall.py
```diff
@@ -231,9 +231,8 @@ class Demo(HasTraits):
         handler=DemoHandler,
     )
 
-    def __init__(self, **traits):
-        super(Demo, self).__init__(**traits)
-        self.plot = _create_plot_component(self.controller)
+    def _plot_default(self):
```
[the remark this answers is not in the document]
aaronayres35: I ended up just throwing the changes into this PR

examples/demo/updating_plot/updating_plot2.py
aaronayres35: I've opened #715

examples/demo/basic/image_from_file.py
```diff
@@ -158,8 +160,8 @@ def _load(self):
         self.pd.set_data("imagedata", image._data)
 
         # Set the title and redraw
-        self.plot.title = os.path.basename(self._load_file)
-        self.plot.request_redraw()
+        plot.title = os.path.basename(self._load_file)
+        plot.request_redraw()
```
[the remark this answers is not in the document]
aaronayres35: I've opened #716

examples/demo/canvas/cliptest.py
aaronayres35: I've closes #576.  I had it "working" just by making the same switch to use `_create_component`, but more will needed to truly fix up the whole example by rewriting it to avoid use of the custom multitouch support / at least make the functionality more useable with a mouse

Issue #621, opened 2021-04-05
Move or remove misc files sitting in docs


rahulporuri:
- [ ] https://github.com/enthought/chaco/blob/master/docs/CHANGES.txt
- [ ] https://github.com/enthought/chaco/blob/master/docs/chaco_layout.txt
- [ ] https://github.com/enthought/chaco/blob/master/docs/datamodel_hierarchy.txt
- [ ] https://github.com/enthought/chaco/blob/master/docs/event_handling.txt
- [ ] https://github.com/enthought/chaco/blob/master/docs/introduction.txt
- [ ] https://github.com/enthought/chaco/blob/master/docs/scipy08_tutorial.keynote.tgz

We can probably extract meaningful information from these files - and we could potentially remove the Doc files and move the pdf files into an archive folder.

- [ ] https://github.com/enthought/chaco/blob/master/docs/Chaco_2_Design.doc and https://github.com/enthought/chaco/blob/master/docs/Chaco_2_Design.pdf
- [ ] https://github.com/enthought/chaco/blob/master/docs/datamodel.doc and https://github.com/enthought/chaco/blob/master/docs/datamodel.pdf

PR #710, opened 2021-04-21
Concatenate change logs

aaronayres35:
Part of #621 

One change log ended just before 4.4 (I think - it atleast goes up to 4.1.0 plus some more changes mentioned after), the other started at 4.4, so I simply stacked them and deleted the older file.


rahulporuri (approved):
LGTM

PR #711, opened 2021-04-21
remove a couple redundant old files

aaronayres35:
Part of #621 

These files are old / redundant with info that already exists in the docs.
The other txt files that remain I did not delete yet because I thought it was more likely we might want to pull some info from them.  These I think can just go.


rahulporuri (approved):
LGTM

PR #669, opened 2021-04-15
Remove references to chaco.shell from the documentation

aaronayres35:
Along with #668, this PR closes #463

This PR removes any lingering mention of `chaco.shell` from the documentation.  In doing so it removes the entire tutorial with ipython, and other smaller portions of the docs describing use of `chaco.shell`

PR #712, opened 2021-04-22
Remove unused related images from docs

rahulporuri:
The doc pages which used these images waere removed in the PRs #668 and #669 but we forgot to remove the images in the PRs. This PR also removes a copy of `simple_line.png` and instead just makes use of the one copy everywhere. This PR also removes an xcf file from the docs and im not even sure who/what uses the file.


aaronayres35 (approved):
LGTM

PR #714, opened 2021-04-22
Unskip no longer failing test

aaronayres35:
This was fixed in #672 but I originally thought it needed https://github.com/enthought/enable/pull/782 in a released version of enable before the test could be unskipped.  

This is actually not the case (although the fix n enable is still needed, the changes in 672 avoid the problem ever getting reached.

This PR simply un skips the test that is now passing


rahulporuri (approved):
LGTM. Did we create an issue to remember to unskip this after an enable release?

aaronayres35:
> LGTM. Did we create an issue to remember to unskip this after an enable release?

I think #232 was the only related issue

Issue #587, opened 2021-03-30
Documentation Issue List


aaronayres35:
This issue is meant to keep a laundry list of all the things we want to address regarding the documentation.  Feel free to add bullets as you see fit.  We can also open individual issues for these things and ultimately close this issue in favor of the standalone issues.  For now I figured it can be used to see everything in one place

- [ ] https://github.com/enthought/chaco/blob/master/docs/source/user_manual/quickstart.rst image links are broken (also top 2 images should maybe be preceded by a sentence or 2.
- [ ] https://github.com/enthought/chaco/blob/master/docs/source/user_manual/modules_and_classes.rst still has some empty sections
- [ ] https://github.com/enthought/chaco/blob/master/docs/source/user_manual/basic_elements/tools.rst is basically stubs only
- [ ] https://github.com/enthought/chaco/blob/master/docs/source/user_manual/basic_elements/overlays.rst is also incomplete with many empty sections
- [ ] https://github.com/enthought/chaco/blob/master/docs/source/user_manual/common_patterns.rst is stubs only
- [ ] https://github.com/enthought/chaco/blob/master/docs/source/user_manual/faq.rst page only has 2 questions. Surely there are more that would be relevant. Also the answers to "What are the pros and cons of Chaco vs. matplotlib?" are from 2007
- [ ] In the user guide I don't see an option to click "Next Topic" like there is for other ets docs. I have to return to the TOC to progress through the user manual (This would remove the need for the links at the bottom of the introduction page as well)
- [ ] Do we want to change the User Guide order? 
- [ ] The content under "Modular design and extensible classes" is repeated in the Chaco section here https://github.com/enthought/chaco/blob/master/docs/source/developer_reference/architecture_overview.rst

Other already existing standalone issues:
- [ ] #586 
- [ ] #530 
- [ ] #305

PR #722, opened 2021-04-28
Link to "how do i", not "faq", from home page

aaronayres35:
part of #587 

The FAQ page does not have very useful information, and it makes more sense to link to the How do I page from the home page of the docs (although it is still a WIP)


rahulporuri (approved):
LGTM

Issue #550, opened 2020-12-21
AbstractPlotRenderers return different array shapes when passing a single point to map_screen


mdsmarte:
`Base2DPlot` and `BaseXYPlot` both implement the `AbstractPlotRenderer` interface, in particular the `map_screen` method.

When a single `(x, y)` point is passed to `Base2DPlot.map_screen`, the returned array is of the form `[[x, y]]` (nested, shape = `(1, 2)`), as set by `GridMapper.map_screen`:

https://github.com/enthought/chaco/blob/4a6683521cce012b5df4830d64f41518798fc817/chaco/grid_mapper.py#L114-L123

When a single `(x, y)` point is passed to `BaseXYPlot.map_screen`, the returned array is of the form `[x, y]` (flattened, shape = `(2,)`):

https://github.com/enthought/chaco/blob/4a6683521cce012b5df4830d64f41518798fc817/chaco/base_xy_plot.py#L336-L353

This difference forces objects that interact with `AbstractPlotRenderer`s to make an assumption about the behavior of the `map_screen` method. For example, the `DataLabelTool` assumes the `map_screen` return shape matches that of `BaseXYPlot` for a single point (`(2,)`):

https://github.com/enthought/chaco/blob/4a6683521cce012b5df4830d64f41518798fc817/chaco/tools/data_label_tool.py#L53-L54

And therefore this tool throws an exception when used with a `Base2DPlot`:

```
File "<REDACTED>/python3.6/site-packages/chaco/tools/data_label_tool.py", line 53, in drag_start
    pointx, pointy = label.component.map_screen(label.data_point)
ValueError: not enough values to unpack (expected 2, got 1)
```

even though it would otherwise be compatible.

This issue also looks like it applies to the `map_data` method although I have not tested it.

aaronayres35:
#802 and #804 have been merged. This PR can be closed after we go through and audit the code base to see if anything was missed

rahulporuri:
> This PR can be closed after we go through and audit the code base to see if anything was missed

I'm guessing you meant "this issue can be closed". Can you do the audit @aaronayres35 ?

aaronayres35:
So I've gone through the code base and I think it looks like we've got everything.

[...]

Issue #289, opened 2016-02-11
CursorTool does not work with LogMapper for both axes


spescha:
Using LogMapper for both axes

```
mag = LinePlot(index = ArrayDataSource(w, sort_order='ascending'),
                        value = ArrayDataSource(H),
                        color = curve_color,
                        name = curve_name,
                        index_mapper = LogMapper(range=view_mag.index_range),
                        value_mapper = LogMapper(range=view_mag.value_range))
```

leads to the error below, when adding a cursor with following lines

```
cursor1 = CursorTool(mag,
                        drag_button="left",
                        show_value_line=False,
                        show_marker=True,
                        line_width=2.0,
                        color='black')
mag.overlays.append(cursor1)
```

The problem is, that LogMapper.map_screen returns a 1-element array, while LinearMapper returns a scalar value, if only one data point has to be mapped.

Following subclass of LogMapper works for me as a workaround

```
class LogMapperCorr(LogMapper):
    """
    Corrects chaco.log_mapper.LogMapper for the usage with cursors. Cursors need
    a 1D-array iterable to unpack the x and y values and the LogMapper returns
    a 2D-array. Therefore LogMapper.map_screen has to return a scalar, if the length 
    of the data array is 1.
    """

    def __init__(self, *args, **kwargs):
        LogMapper.__init__(self, *args, **kwargs)

    def map_screen(self, data_array):
        arr = LogMapper.map_screen(self, data_array)

        if len(arr)==1:
            return arr[0]
        else:
            return arr
```

```
Traceback (most recent call last):
  File "C:\Users\speschda\AppData\Local\Enthought\Canopy\User\lib\site-packages\enable\qt4\base_window.py", line 198, in paintEvent
    self.handler.paintEvent(event)
  File "C:\Users\speschda\AppData\Local\Enthought\Canopy\User\lib\site-packages\enable\qt4\base_window.py", line 50, in paintEvent
    self._enable_window._paint(event)
  File "C:\Users\speschda\AppData\Local\Enthought\Canopy\User\lib\site-packages\enable\abstract_window.py", line 467, in _paint
    self.component.draw(gc, view_bounds=(0, 0, size[0], size[1]))
  File "C:\Users\speschda\AppData\Local\Enthought\Canopy\User\lib\site-packages\enable\component.py", line 424, in draw
    self._draw(gc, view_bounds, mode)
[...]

Issue #272, opened 2015-05-14
BaseXYPlot creates invalid screen points


cfarrow:
It is possible to get invalid screen points from a `BaseXYPlot.map_screen`. 

Starting [here](https://github.com/enthought/chaco/blob/2fc52518278c52e81a7e07589bb6aeaaca361770/chaco/base_xy_plot.py#L346):
If data_array is 1-dimensional and one of the points is out of bounds, then the `LinearMapper` being used to map the screen point ([here](https://github.com/enthought/chaco/blob/2fc52518278c52e81a7e07589bb6aeaaca361770/chaco/linear_mapper.py#L52)) will return an array. Meanwhile, for the in-bounds point a scalar will be returned ([here](https://github.com/enthought/chaco/blob/2fc52518278c52e81a7e07589bb6aeaaca361770/chaco/linear_mapper.py#L54)). Then the `BaseXYPlot` packs the array and scalar into a single object array which is not the right thing to do.

I think the right solution is to have the BaseXYPlot handle this corner case, but the `LinearMapper` might be to blame since in both places the proper behavior is not documented.

PR #724, opened 2021-05-03
[WIP] Audit return types

aaronayres35:
In this PR I try to make a bunch of changes at once to input / return types across chaco and see what happens.

Note the changes in this PR could fix (potentially) #272, #289, #550, and more.  Eventually I will verify these things and try to write regression tests. 

I currently am treating this PR as an experiment.  I imagine the changes here will likely be pulled out into various, more incremental, small PRs that slowly make a migration.  If this PR ever is to be merged in its entirety, it will require substantial clean up and tests.

A struggle with making these changes is lack of coverage (we could be breaking other things internally even if test suite passes), and Hyrum's law.  Almost certainly downstream users are depending on these return types in one way or the other.  The issue description on #550 very nicely describes how these differences across the codebase force other code to assume one way or the other... Users likely have had to make that assumption that holds true for their particular use case.

### DIFF
--- a/chaco/barplot.py
+++ b/chaco/barplot.py
@@ -165,9 +165,17 @@ class BarPlot(AbstractPlotRenderer):
         an array.
 
         Implements the AbstractPlotRenderer interface.
+
+        Parameters
+        ----------
+        data_array: array of shape (N, 2)
+
+        Returns
+        -------
+        array of shape (N, 2)
         """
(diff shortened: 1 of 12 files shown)


aaronayres35:
I am going to close this now as I do not intend for it to ever be merged, and #726 exists as a first step

PR #725, opened 2021-05-04
Use "super" instead of calling method on class explicitly

rahulporuri:
Instead of `LassoOverlay._draw_component`, we use `super()._draw_component` instead. `LassoOverlay` is what `RegressionOverlay` inherits from.


aaronayres35 (approved):
LGTM

Issue #542, opened 2020-11-20
RangeSelection2D not working as expected


sogoogos:
**Problem Description**
I am trying to add the functionality to set ROI with a mouse using `RangeSelection2D`.
However, I am getting the following error after right-clicking the image displayed.  

```
File "/Users/sshiozawa/.edm/envs/python-class/lib/python3.6/site-packages/chaco/linear_mapper.py", line 67, in map_data
    return (screen_val - self.low_pos) / self._scale + self.range.low
TypeError: unsupported operand type(s) for -: 'list' and 'float'
```

I see some old attempt [here](https://github.com/enthought/chaco/pull/162).

**Reproduction Steps:**
Run the following code with an image file and then right-click the image in the interface. 

```python
from skimage import io
import matplotlib.pyplot as plt
from traits.api import Array, Bool, Button, Enum, File, Float, HasTraits, Instance, Int, List, Str, on_trait_change
from traitsui.api import CheckListEditor, Group, Item, HGroup, spring, UItem, View
from chaco.api import Plot, ArrayPlotData
from chaco.tools.api import PanTool, RangeSelection2D, ZoomTool
from enable.api import BaseTool, ComponentEditor
class Image_GUI(HasTraits):
    # Define Parameters
    file_path = File(exists=True)
    img = Array()
    plot = Instance(Plot)
    # Define File Select Function
    def _file_path_changed(self):
        self.img = io.imread(self.file_path)
    # Define Plot Function
    def _plot_default(self):
        img = self.img
        # Show Plot if Image is loaded
        if len(img) > 0:
            # Show Plot
            data = ArrayPlotData(img=img)
            plot = Plot(data)
            plot.img_plot('img', origin='top left')
            # Set Aspect Ratio
            height, width, _ = img.shape
            plot.aspect_ratio = width / height
            plot.tools.append(RangeSelection2D(plot))
        else:
            plot = Plot()
        return plot
    # Run Plot Method if file is changed
    @on_trait_change('file_path')
    def create_new_plot(self):
        self.plot = self._plot_default()
    # Configure GUI Settings
    traits_view = View(Item('file_path'),
                       UItem('plot', editor=ComponentEditor()))
if __name__ == '__main__':
    img_gui = Image_GUI()
    img_gui.configure_traits()
```

**Expected behavior:**

[Describe expected behavior here]

**OS, Python version:** 
macOS, Python 3.6.12

rahulporuri:
[...]

PR #726, opened 2021-05-04
Cleanup map_screen return types

### LINKED ISSUE #272: BaseXYPlot creates invalid screen points

aaronayres35:
fixes #272, #289, #550

This PR goes through the full chaco code base and makes sure definitions of `map_screen` always return ndarrays of a consistent shape.  
~To do so, it also converts inputs to an ndarray of an appropriate shape.  For example `BaseXYPlot.map_screen` should have 2-dim data points, so it converts its input into Nx2 array.~
~This caused some problem with segment plot which actually keeps its data as Nx2x2 array where a "data point" is really 2 points representing a segment.  To address this I had SegmentPlot overwrite its map_screen method.~

~Currently the first commit has all the "real" changes, and it may be easier for a reviewer to look there specifically first.  The second commit simply makes clean up changes to use numpy as np everywhere.  I had been doing so in my changes, but the original code was importing specific things from numpy.  I made them consistent by just using np everywhere.~

~This PR is a WIP as it still needs to add extensive tests~ I've added various tests

Similar problems exist with `map_data`, for example see: #542 

This PR does a couple things.  First, in the case where the given input to `map_screen` is empty, it will now return an empty ndarray or appropriate shape rather than an empty list.  Additionally, inputs to map_screen are checked / forced to be the arrays of correct shape.  In some cases this is Nx2, others it is Nx2x2, others it is Nx1.  These shapes were commented on before (and at times assumed) but not enforced.  This change prevents strange bugs that would result from passing in a single point which would have shape (2,) and then calling transpose and unpacking the result leading to working with individual floating point values rather than singleton arrays. As per comment below, BaseXYPlot now converts the data to an array, checks if it has ndim == 1 in which case it reshapes it to (-1,2) (handling the case just described), and then checks if the shape ends with 2. If it does not a Value Error is raised as the ensuing calculations assume this to be the case.  
In other cases we can be strict in expecting Nx2 or Nx2x2 for example, eg Barplot or SegmentPlot.


corranwebster:
I think this may be a bit strict on the expected shapes, but my brain isn't in chaco mode right now.  For example, I seem to recall that the implicit assumption of a BaseXYPlot map method is that the shape is (..., 2) (ie. we only make assumptions about the size of the last dimension.

chaco/segment_plot.py
```diff
@@ -132,6 +132,28 @@ def hittest(self, *args, **kwargs):
...
+
+        x_ary, y_ary = np.transpose(data_array)
+
+        sx = self.index_mapper.map_screen(x_ary)
+        sy = self.value_mapper.map_screen(y_ary)
+        if self.orientation == "h":
+            return np.transpose(np.array((sx, sy)))
+        else:
+            return np.transpose(np.array((sy, sx)))
+
```
aaronayres35, line 156: Note this is effectively exactly the same as BaseXYPlot.map_screen only here we enforce Nx2x2.  Previously this worked with BaseXYPlot.map_screen as the `reshape` line wasn't there to enforce Nx2 (hence Corran's comment).  However, in the `BaseXYPlot` there was a comment explicitly saying "data array is Nx2 array" but that wasn't actually what the code assumed.

To avoid potential breaks we probably don't want to make that change to `BaseXYPlot`.
But I don't see a problem with being more specific on subclasses.  ie I think it makes sense for SegmentPlot to enforce an Nx2x2 data array? That's what it wants to be working with anyway.

`BaseXYPlot` should also be made to be explicit about the input shape.  Even if it is (..., 2) that should be clearly documented / enforced.

aaronayres35:
I've pulled out parts of this PR into standalone PRs #802 #803 #804

I will eventually open 1 last PR to pull out all the more general changes of this PR which weren't specific to an issue, but rather were about general consistency across the codebase.  Once that is opened and the new PRs merged we can close this PR.

aaronayres35:
This PR shouldn't be merged.  I merged master into this branch basically just to check what changes still need to be pulled out into stand alone PRs.

chaco/grid_mapper.py
```diff
@@ -126,6 +126,10 @@ def map_screen(self, data_pts):
 
         Maps values from data space into screen space.
         """
+        # ensure data_array is an Nx2 ndarray
+        data_pts = array(data_pts)
+        data_pts = data_pts.reshape(-1, 2)
```
aaronayres35, line 131: the `column_stack` call below actually avoids any errors here making this code un-necessary.  Ref:
https://github.com/enthought/chaco/blob/d83e4b8db5bf44be40c450b364ebd159a00c4d62/chaco/tests/test_grid_mapper.py#L36-L42

chaco/data_view.py
```diff
@@ -242,7 +242,11 @@ def map_screen(self, data_array):
         """Maps an array of data points to screen space and returns an array
         of screen space points.
         """
-        # data_array is Nx2 array
+
+        # ensure data_array is an Nx2 ndarray
+        data_array = array(data_array)
+        data_array = data_array.reshape(-1,2)
```
aaronayres35, line 248: these are actually redundant / unneeded as in the return statement we call `array`

aaronayres35:
Closing as the content of this PR has been pulled out into smaller PRs

Issue #47, opened 2012-02-16
PolarMapper does not map to polar coordinates


pberkes:
At the moment PolarMapper is the same as LinearMapper, and one needs to map from polar to catesian coordinates by hand.

See: docs in polar_mapper.py, create_polar_plot in plot_factory.py

PR #732, opened 2021-05-07
re-export PolarLineRenderer and PolarMapper from chaco.api

### LINKED ISSUE #629: Re-export PolarLineRenderer and PolarMapper from chaco.api
Ref [`PolarLineRenderer`](https://github.com/enthought/chaco/blob/c758a65222a2caef9a3d2c074ae9dd5775a3a77e/chaco/polar_line_renderer.py), [`PolarMapper`](https://github.com/enthought/chaco/blob/c758a65222a2caef9a3d2c074ae9dd5775a3a77e/chaco/polar_mapper.py)

aaronayres35:
closes #629 

This PR simply re-exports the 2 classes from the api module.

It also pulls a comment about `PolarMapper` currently just being LinearMapper into the docstring so it appears on api docs.  Eventually we will want to update it, see #47.  In the meanwhile it is fine to like in the api as long as this is documented.
(Actually looking at it, it looks like some changes have been made to LinearMapper that weren't copied over... I may go ahead and pull those directly so that the PolarMapper really does just match the LinearMapper like it says).
EDIT: done. The PolarMapper is now the current LinearMapper rather than an old version of the LinearMapper.

### DIFF
--- a/chaco/polar_mapper.py
+++ b/chaco/polar_mapper.py
@@ -12,11 +12,11 @@ from traits.api import Bool, Float
 # Local relative imports
 from .abstract_mapper import AbstractMapper
 
-###############################################################
-# same as linear mapper at the moment... to be modified later #
-###############################################################
+
 class PolarMapper(AbstractMapper):
     """
+    Same as linear mapper at the moment... to be modified later
+
     Maps a 1-D data space to and from screen space by specifying a range in
     data space and a corresponding fixed line in screen space.
 
--- a/chaco/polar_mapper.py
+++ b/chaco/polar_mapper.py
@@ -4,7 +4,7 @@ into a 1-D output space.
 """
 
 # Major library imports
-from numpy import array
+from numpy import array, float64, full_like, ndarray
 
 # Enthought library imports
 from traits.api import Bool, Float
@@ -29,8 +29,11 @@ class PolarMapper(AbstractMapper):
     # Private traits
     # ------------------------------------------------------------------------
 
-    _scale = Float(1.0)  # number of screen space units per data space unit
+    # Number of screen space units per data space unit.
+    _scale = Float(1.0)
+    # Is the range of the screen space empty?
     _null_screen_range = Bool(False)
+    # Is the range of the data space empty?
     _null_data_range = Bool(False)
 
     # ------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


chaco/polar_mapper.py
```diff
@@ -29,8 +29,11 @@ class PolarMapper(AbstractMapper):
     # Private traits
     # ------------------------------------------------------------------------
 
-    _scale = Float(1.0)  # number of screen space units per data space unit
+    # Number of screen space units per data space unit.
```
rahulporuri, line 32:
```suggestion
    #: Number of screen space units per data space unit.
```

chaco/polar_mapper.py
```diff
@@ -29,8 +29,11 @@ class PolarMapper(AbstractMapper):
...
+    _scale = Float(1.0)
+    # Is the range of the screen space empty?
```
rahulporuri, line 34:
```suggestion
    #: Is the range of the screen space empty?
```

chaco/polar_mapper.py
```diff
@@ -29,8 +29,11 @@ class PolarMapper(AbstractMapper):
...
     _null_screen_range = Bool(False)
+    # Is the range of the data space empty?
```
rahulporuri, line 36:
```suggestion
    #: Is the range of the data space empty?
```

chaco/polar_mapper.py
```diff
@@ -40,27 +43,38 @@ class PolarMapper(AbstractMapper):
...
-        values from data space into screen space.
+        Overrides AbstractMapper. Maps values from data space into screen space.
         """
         self._compute_scale()
         if self._null_data_range:
-            return array([self.low_pos] * len(data_array))
+            if isinstance(data_array, (tuple, list, ndarray)):
+                return full_like(data_array, self.low_pos, dtype=float64)
+            else:
+                return array([self.low_pos])
```
rahulporuri, line 53: I'm not sure why these changes were made. Were these changes copied over from the linear mapper?

rahulporuri: I have the same comment about the changes in `map_data` and `_compute_scale`

chaco/polar_mapper.py
aaronayres35: This are privately named traits, but I agree they do seem useful to have in the api docs. I will go ahead with the # to #: changes

aaronayres35: Actually, I'm having second thoughts.  I don't think users should need to ever touch these traits correct?

rahulporuri (approved):
LGTM. I haven't tested to check that the example still works.

PR #735, opened 2021-05-11
Change default padding based on axis titles

### LINKED ISSUE #564: Y label gets clipped
**Problem Description**

The Y label in the following example doesn't get displayed while the X label gets displayed.

why-chaco-why

**Reproduction Steps:**

```python
from numpy import linspace, sin
from traitsui.api import View, Item
from chaco.api import Plot, ArrayPlotData

class LinePlot(HasTraits):
    plot = Instance(Plot)

    traits_view = View(
        Item('plot',editor=ComponentEditor(), show_label=False),
        width=500, height=500, resizable=True, title="Chaco Plot")

    def _plot_default(self):
        x = linspace(-14, 14, 100)
        y = sin(x) * x**3
        plotdata = ArrayPlotData(x=x, y=y)

        plot = Plot(plotdata)
        plot.plot(("x", "y"), type="line", color="blue")
        plot.title = "sin(x) * x^3"
       
        plot.x_axis.title = "X AXIS"
        plot.y_axis.title = "Y AXIS"
       
        return plot

if __name__ == "__main__":
    LinePlot().configure_traits()
```

**Expected behavior:**

We expect the Y label to be displayed.

**OS, Python version:** Windows 10 Pro. Python 3.6.0. Chaco 4.8.0. Enable 4.8.1.

Originally reported on the ETS Users Google Group - https://groups.google.com/g/ets-users/c/gHQmFsVMVhw

aaronayres35:
closes #564

Currently, with manual testing this seems to work how I want it to but I am not sure the current solution is worth the added complexity. 
If we're going to add complexity to the code I think it might be better to try and get a more "perfect" approach that sets padding directly based off of the axis title extent.  Hence the WIP title as I am still thinking about alternate solutions.

### DIFF
--- a/chaco/data_view.py
+++ b/chaco/data_view.py
@@ -257,16 +257,16 @@ class DataView(OverlayPlotContainer):
         return self._find_padding("right")
     
     def _set_padding_top(self, value):
-        self._padding_left = value
+        self._padding_top = value
 
     def _set_padding_bottom(self, value):
-        self._padding_left = value
+        self._padding_bottom = value
 
     def _set_padding_left(self, value):
         self._padding_left = value
 
     def _set_padding_right(self, value):
-        self._padding_left = value
+        self._padding_right = value
 
     border_visible = True
 
--- a/chaco/data_view.py
+++ b/chaco/data_view.py
@@ -3,7 +3,7 @@ functions.
 """
 from numpy import array, transpose
 
-from traits.api import Bool, Enum, Instance, Int, observe, Property
+from traits.api import Bool, Enum, Instance, Int, observe, Property, Union
 from enable.api import color_table
 
 from .abstract_overlay import AbstractOverlay
@@ -221,10 +221,10 @@ class DataView(OverlayPlotContainer):
         observe='y_axis.[title,orientation], x_axis.[title,orientation]'
     )
 
-    _padding_top = Int(0)
-    _padding_bottom= Int(0)
-    _padding_left = Int(0)
-    _padding_right = Int(0)
+    _padding_top = Union(None, Int())
+    _padding_bottom= Union(None, Int())
+    _padding_left = Union(None, Int())
+    _padding_right = Union(None, Int())
 
     def _find_padding(self, side):
         MAP = {
@@ -233,7 +233,7 @@ class DataView(OverlayPlotContainer):
             "left": "_padding_left",
             "right": "_padding_right",
         }
-        if getattr(self, MAP[side]):
+        if getattr(self, MAP[side]) is not None:
             return getattr(self, MAP[side])
         else:
             if self.y_axis:
--- a/chaco/data_view.py
+++ b/chaco/data_view.py
@@ -226,53 +226,35 @@ class DataView(OverlayPlotContainer):
     _padding_left = Int(0)
     _padding_right = Int(0)
 
-    def _get_padding_top(self):
-        if self._padding_top:
-            return self._padding_top
+    def _find_padding(self, side):
+        MAP = {
+            "top": "_padding_top",
+            "bottom": "_padding_bottom",
+            "left": "_padding_left",
+            "right": "_padding_right",
+        }
+        if getattr(self, MAP[side]):
+            return getattr(self, MAP[side])
         else:
             if self.y_axis:
-                if (self.y_axis.title and self.y_axis.orientation == "top"):
+                if (self.y_axis.title and self.y_axis.orientation == side):
                     return 80
             if self.x_axis:
-                if (self.x_axis.title and self.x_axis.orientation == "top"):
+                if (self.x_axis.title and self.x_axis.orientation == side):
                     return 80
             return 50
+
+    def _get_padding_top(self):
+        return self._find_padding("top")
 
     def _get_padding_bottom(self):
-        if self._padding_bottom:
-            return self._padding_bottom
-        else:
-            if self.y_axis:
-                if (self.y_axis.title and self.y_axis.orientation == "bottom"):
-                    return 80
-            if self.x_axis:
-                if (self.x_axis.title and self.x_axis.orientation == "bottom"):
-                    return 80
-            return 50
+        return self._find_padding("bottom")
 
     def _get_padding_left(self):
-        if self._padding_left:
-            return self._padding_left
-        else:
-            if self.y_axis:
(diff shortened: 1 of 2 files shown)


chaco/data_view.py
```diff
@@ -208,10 +208,65 @@ class DataView(OverlayPlotContainer):
     bgcolor = "white"
 
     #: Padding defaults.
-    padding_top = 50
-    padding_bottom = 50
-    padding_left = 50
-    padding_right = 50
+    padding_top = Property(
```
aaronayres35, line 211: if explicitly set (or if `padding` trait is sett, which then explicitly sets these), that value will always be used.  Setting sets the shadow traits (with leading underscore), and in the getters we explicitly check if the shadow trait is set and return it.  Otherwise we then check for axis titles and adjust as needed either returning 80 or 50 as a default.

Note I am just now considering if a user explicitly sets one off these traits to 0 I dont think it will work as then we will not enter the `if getattr(self, MAP[side])` block.  This is a bug that needs fixing.

chaco/data_view.py
```diff
@@ -208,10 +208,65 @@ class DataView(OverlayPlotContainer):
...
+    )
+    padding_bottom = Property(
+        observe='y_axis.[title,orientation], x_axis.[title,orientation]'
+    )
+    padding_left = Property(
+        observe='y_axis.[title,orientation], x_axis.[title,orientation]'
+    )
+    padding_right = Property(
+        observe='y_axis.[title,orientation], x_axis.[title,orientation]'
+    )
```
aaronayres35, line 222: I handle all 4 cases because it is possible to set axis orientation to "top" or "bottom" as well

aaronayres35: as an example, this code:
```
rom numpy import linspace, sin
from traitsui.api import View, Item
from chaco.api import Plot, ArrayPlotData

class LinePlot(HasTraits):
    plot = Instance(Plot)

    traits_view = View(
        Item('plot',editor=ComponentEditor(), show_label=False),
        width=500, height=500, resizable=True, title="Chaco Plot")

    def _plot_default(self):
        x = linspace(-14, 14, 100)
        y = sin(x) * x**3
        plotdata = ArrayPlotData(x=x, y=y)

        plot = Plot(plotdata)
        plot.plot(("x", "y"), type="line", color="blue")
        plot.title = "sin(x) * x^3"

        plot.padding_left = 0
        
        plot.x_axis.title = "X AXIS"
        plot.y_axis.title = "Y AXIS"
        
        plot.y_axis.orientation = "top"
       
        return plot

if __name__ == "__main__":
    LinePlot().configure_traits()
```

yields this on master:
Screen Shot 2021-05-12 at 9 38 47 AM

and this on this branch:
Screen Shot 2021-05-12 at 9 39 02 AM

aaronayres35: The extra padding on the bottom on this branch (80 rather than 50) I'm not a huge fan of.  This is due to the simplistic nature of just adding space if we have an axis title.  As you can see, at the top we need extra space so axis title doesn't overlap with plot title.  Similarly left right we need it (hence the original issue), because axis ticks / values take up some of the spacing and sort of push the axis title further away from the plot.  The x axis doesn't usually have this problem because of the orientation of the numbers on the ticks.  In other words the text for `1000` for example is much wider than it is tall, and `100000` even wider, but the same height.

In reality an ideal solution would calculate explicitly the extent of the ticks, and the size of the axis title and allocate spacing to perfectly fit what we need.  The awkward thing about this code feels like it would live in a draw method inside the axis itself, but the DataView is where the padding is held.  Should the Axis take control over the DataViews padding?  That feels wrong.
In any case that is a separate issue / approach entirely.  I am just pointing out something I don't love about the current solution.

Perhaps we just make the assumption that padding_bottom can be 50 regardless of the presence or absence of an axis title?  We were just assuming 50 always before anyway.

chaco/data_view.py
```diff
@@ -208,10 +208,65 @@ class DataView(OverlayPlotContainer):
...
+
+    _padding_top = Union(None, Int())
+    _padding_bottom= Union(None, Int())
+    _padding_left = Union(None, Int())
+    _padding_right = Union(None, Int())
```
aaronayres35, line 227: @rahulporuri 
I looked into it and `Int(allow_none=True)` does not work / is not implemented. 

```
>>> from traits.api import HasTraits, Instance, Int
>>> class Thing(HasTraits):
...     a = Int()
...     b = Int(allow_none=True)
...     c = Instance(HasTraits, allow_none=False)
...     d = Instance(HasTraits)  # default value of allow_none for Instance trait type is True
... 
>>> thing = Thing()
>>> thing.d = None
>>> thing.c = None
  File "<stdin>", line 1, in <module>
  File "/Users/aayres/Desktop/traits/traits/base_trait_handler.py", line 75, in error
    object, name, self.full_info(object, name, value), value
traits.trait_errors.TraitError: The 'c' trait of a Thing instance must be a HasTraits, but a value of None <class 'NoneType'> was specified.
>>> thing.b = None
  File "<stdin>", line 1, in <module>
  File "/Users/aayres/Desktop/traits/traits/base_trait_handler.py", line 75, in error
    object, name, self.full_info(object, name, value), value
traits.trait_errors.TraitError: The 'b' trait of a Thing instance must be an integer, but a value of None <class 'NoneType'> was specified.
```

rahulporuri: I think this deserves an issue in traits - traits should explicitly complain during class definition instead of at runtime.

aaronayres35: That makes sense, for my own understanding we want to catch a user setting `allow_none` in a trait definition correct?  I suspect, since `allow_none` is not a supported arg/kwarg for the `Int` trait type, it is now being used as metadata? Ie one could do something like `@observe('+allow_none')`?
We should probably special handle this case

rahulporuri (approved):
LGTM.

I've been putting off reviewing this PR because proposed solution looks slightly overworked/unweidly and I was hoping i'd be able to conjure something better out of thin air.

Alas, some solution is better than no solution so let's go ahead with this.

A couple of comments though :
- it would be good to add tests for the new code - to check that the padding values are what we expect
- i'm not sure where but i would like a reference to the original issue in the new code so that people who come asking questions know where to look.

aaronayres35:
@rahulporuri this PR look good to push through?

rahulporuri:
still LGTM

Issue #736, opened 2021-05-12
Avoid use of soon to be removed deprecated _draw_component


aaronayres35:
Currently in various places in the code base we are using deprecated methods from enable.
Ref: https://github.com/enthought/enable/pull/814 and https://github.com/enthought/enable/issues/794

The only uses of `_draw_component` that actually end up calling the methods in enable (ie would break if removed in enable) are mentioned here https://github.com/enthought/enable/issues/781#issuecomment-831183408
(note for other uses the class explicitly overrides `_draw_component` and doesn't call super)
The other uses we still likely want to remove at some point as the method is intended for compatibility with the "old" draw approach.

The other methods removed in the linked PR are not used in chaco.

PR #737, opened 2021-05-13
remove calls to Component._draw_component

aaronayres35:
Part of #736 (maybe this is enough to close the issue?  We may want to remove all use of _draw_component, even uses that won't break)

These 2 calls actually were calling `Component._draw_component` living in `enable` which currently is a no-op:
https://github.com/enthought/enable/blob/21c299e1229d74e9db701487a613934e332a7225/enable/component.py#L344-L350

There are other uses of `_draw_component` across chaco that can be removed, but they will not cause a break if `_draw_component` is removed in `enable` (ref https://github.com/enthought/enable/pull/814), because the method gets overwritten in chaco.

### DIFF
--- a/chaco/layers/status_layer.py
+++ b/chaco/layers/status_layer.py
@@ -114,8 +114,6 @@ class StatusLayer(AbstractOverlay):
 
             self.document.render(gc)
 
-            self._draw_component(gc, view_bounds, mode)
-
     def fade_out(self):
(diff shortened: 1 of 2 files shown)


rahulporuri (approved):
LGTM.

This isn't enough to close #736 as we also want to remove the "deprecated" comments in the `_draw_component` once we take a better look at those methods.

PR #740, opened 2021-05-17
use mock through unittest

aaronayres35:
`mock` is now in the standard library since python 3.3. This PR updates the codebase to use mock via unittest.

Python 2 is end of life and we plan to drop support for 5.0.0 release


rahulporuri (approved):
LGTM

PR #742, opened 2021-05-18
Replace bg_color with bgcolor

aaronayres35:
ref: https://github.com/enthought/enable/pull/816#pullrequestreview-661608507

This PR removes uses of the soon to be removed, deprecated `bg_color` trait from enable.  For consistency it renames all uses of `bg_color` in the codebase with `bgcolor` (even just temp variables), in order to keep consistent.  Note, all changes occur in example code

### DIFF
--- a/examples/demo/basic/contour_cmap_plot.py
+++ b/examples/demo/basic/contour_cmap_plot.py
@@ -62,7 +62,7 @@ def _create_plot_component():
     # Tweak some of the plot properties
     lplot.title = "Colormap and contours"
     lplot.padding = 20
-    lplot.bg_color = "white"
+    lplot.bgcolor = "white"
     lplot.fill_padding = True
 
     # Add some tools to the plot
@@ -110,7 +110,7 @@ def _create_plot_component():
     # Tweak some of the plot properties
     rplot.title = "Varying contour lines"
     rplot.padding = 20
-    rplot.bg_color = "white"
+    rplot.bgcolor = "white"
     rplot.fill_padding = True
 
     # Create a container and add our plots
@@ -127,7 +127,7 @@ def _create_plot_component():
 # Attributes to use for the plot view.
 size = (950, 650)
 title = "Some contour plots"
-bg_color = "lightgray"
+bgcolor = "lightgray"
 
 # ===============================================================================
 # # Demo class that is used by the demo.py application.
@@ -139,7 +139,7 @@ class Demo(HasTraits):
         Group(
             Item(
                 "plot",
-                editor=ComponentEditor(size=size, bgcolor=bg_color),
+                editor=ComponentEditor(size=size, bgcolor=bgcolor),
                 show_label=False,
             ),
             orientation="vertical",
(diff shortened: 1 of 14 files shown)


examples/demo/basic/contour_cmap_plot.py
```diff
@@ -62,7 +62,7 @@ def _create_plot_component():
     # Tweak some of the plot properties
     lplot.title = "Colormap and contours"
     lplot.padding = 20
-    lplot.bg_color = "white"
+    lplot.bgcolor = "white"
```
rahulporuri, line 65: Looks like this was just plain broken earlier. `bgcolor` is a trait on `DataView`, which `Plot` inherits from. Note that `DataView` inherits from enable `Component`, which also defines the `bgcolor` trait.

examples/demo/basic/contour_cmap_plot.py
```diff
@@ -110,7 +110,7 @@ def _create_plot_component():
     # Tweak some of the plot properties
     rplot.title = "Varying contour lines"
     rplot.padding = 20
-    rplot.bg_color = "white"
+    rplot.bgcolor = "white"
```
rahulporuri, line 113: same comment as above - `bg_color` trait doesn't even exist on a `Plot`.

examples/demo/basic/image_plot.py
```diff
@@ -79,7 +79,7 @@ class Demo(HasTraits):
         Group(
             Item(
                 "plot",
-                editor=ComponentEditor(size=size, bgcolor=bg_color),
+                editor=ComponentEditor(size=size, bgcolor=bgcolor),
```
rahulporuri, line 82: FTR : `ComponentEditor` is the only place where I see someone set `bgcolor` on an enable `AbstractWindow` - and it sets `bgcolor`, not the deprecated `bg_color`.

examples/demo/basic/scatter_rect_select.py
```diff
@@ -96,7 +96,6 @@ def _create_plot_component():
 # Attributes to use for the plot view.
 size = (650, 650)
 title = "Scatter plot with selection"
-bg_color = "lightgray"
```
rahulporuri, line 99: why not pass this to the `ComponentEditor` - like what we do in the other examples

examples/demo/basic/scatter_select.py
```diff
@@ -87,7 +87,6 @@ def _create_plot_component():
 # Attributes to use for the plot view.
 size = (650, 650)
 title = "Scatter plot with selection"
-bg_color = "lightgray"
```
rahulporuri, line 90: same comment as above - why not pass this to the `ComponentEditor`, like we do with the other examples

rahulporuri (approved):
LGTM with a couple of comments/recommendations

rahulporuri:
Here's the original commit where `bg_color` was deprecated - https://github.com/enthought/enable/commit/1f5e90f241ea61b1cedf2026a022cca509df9301

examples/demo/basic/scatter_rect_select.py
aaronayres35: I can add that here in this PR.  I deleted it just because previously this `bg_color` variable was just unused entirely.

It makes sense to actually use it though

PR #743, opened 2021-05-18
Remove all definitions / uses of _draw_component

### LINKED ISSUE #736: Avoid use of soon to be removed deprecated _draw_component

aaronayres35:
closes #736 

Note that `_draw_component` has been removed from enable in https://github.com/enthought/enable/pull/814
It was a deprecated method that was a part of the old rendering approach.

These occurrences were all cases in which `_draw_component` had been overwritten in chaco and called in a few places.  This was intended to preserve backwards compatibility, but is now no longer necessary. The logic can simply be copied over into the appropriate methods (which previously called `_draw_component`).


chaco/axis.py
```diff
@@ -244,23 +244,38 @@ def overlay(self, component, gc, view_bounds=None, mode="normal"):
...
+                )
+            if self.title:
+                self._draw_title(gc)
+
+            self._draw_ticks(gc)
+            self._draw_labels(gc)
+
+        self._cache_valid = True
 
     def _draw_overlay(self, gc, view_bounds=None, mode="normal"):
```
aaronayres35, line 274: `_draw_overlay` and `overlay` are now currently the exact same method.  Perhaps one should just call the other to avoid the duplication.  I think it would make more sense to have `overlay` call `_draw_overlay`.
AFAIK we need to keep both methods, but this should be investigated more closely.

chaco/base_xy_plot.py
```diff
@@ -474,12 +474,6 @@ def get_screen_points(self):
 
     def _draw_plot(self, gc, view_bounds=None, mode="normal"):
         """Draws the 'plot' layer."""
-        self._draw_component(gc, view_bounds, mode)
-
-    def _draw_component(self, gc, view_bounds=None, mode="normal"):
-        # This method should be folded into self._draw_plot(), but is here for
-        # backwards compatibilty with non-draw-order stuff.
```
aaronayres35, line 481: This comment effectively sums up the aim of this PR

chaco/tools/regression_lasso.py
```diff
@@ -60,8 +60,8 @@ class RegressionOverlay(LassoOverlay):
         ),
     )
 
-    def _draw_component(self, gc, view_bounds=None, mode="normal"):
-        super()._draw_component(gc, view_bounds, mode)
+    def overlay(self, other_component, gc, view_bounds=None, mode="normal"):
```
aaronayres35, line 63: I tested this by running `examples/demo/basic/regression.py` and everything seems to work as expected

chaco/grid.py
```diff
@@ -333,31 +333,59 @@ def _draw_overlay(self, gc, view_bounds=None, mode="normal"):
...
+
+        Overrides AbstractOverlay.
         """
         # What we're really trying to do with a grid is plot contour lines in
         # the space of the plot.  In a rectangular plot, these will always be
         # straight lines.
         if not self.visible:
             return
-
+        self._compute_ticks(other_component)
```
aaronayres35, line 387: This method now only differs from `_draw_overlay` in this line and the `self._cache_valid = False` line as this is what `overlay` had done differently previously as well

chaco/axis.py
```diff
@@ -244,23 +244,38 @@ def overlay(self, component, gc, view_bounds=None, mode="normal"):
...
         """Draws the overlay layer of a component.
 
         Overrides PlotComponent.
         """
-        self._draw_component(gc, view_bounds, mode)
-
-    def _draw_component(
```
rahulporuri, line 256: I would prefer leaving the `_draw_component` method as it is instead of duplicating it in `overlay` and `_draw_overlay`.

we could simply update the outdated the docstring in this case.

chaco/grid.py
```diff
@@ -333,31 +333,59 @@ def _draw_overlay(self, gc, view_bounds=None, mode="normal"):
...
-        Overrides AbstractOverlay.
-        """
-        if not self.visible:
+        if len(self._tick_positions) == 0:
             return
-        self._compute_ticks(other_component)
-        self._draw_component(gc, view_bounds, mode)
-        self._cache_valid = False
 
-    def _draw_component(self, gc, view_bounds=None, mode="normal"):
```
rahulporuri, line 349: same comment as before regarding the duplication

rahulporuri (commented):
Apart from two instances of duplication, this PR mostly LGTM

chaco/axis.py
rahulporuri: > I think it would make more sense to have overlay call _draw_overlay.

sounds good to me.

> AFAIK we need to keep both methods, but this should be investigated more closely.

I'm not a 100% sure why both methods need to exist. I know that `_draw_overlay` is needed by the new (and currently used) way of drawing the various layers in a chaco plot. Not sure what `overlay` is used for.

chaco/axis.py
rahulporuri: never mind. let's make `overlay` use `_draw_overlay` like you mentioned in the earlier comment

chaco/grid.py
rahulporuri: like before, you can make `overlay` call `_draw_overlay`

aaronayres35: Here they are _very slightly_ different... 😕 
eg `overlay` calls `self._compute_ticks(other_component)` but `_draw_overlay` does not...

chaco/axis.py
aaronayres35: Looking at this PR again I realize that `overlay` and `_draw_overlay` do very slightly different things. I think it is going to be worth the deeper investigation into what both are uesd for / if we can refine things down so only 1 is necessary at all. 
If they are in fact both needed it is possible they will need to exist stand alone with very similar (largely copied) code, but some minor discrepancies.

aaronayres35: Digging into enable:
`overlay` is a method defined on `AbstractOverlay`, and it takes `other_component` as an argument.  Effectively the overlay draws itself on top of another component.

`_draw_overlay` originates from `Component` which keeps track of a list of `AbstractOverlays`, and when `_draw_overlay` is called, it iterates over each of those overlays and calls their respective `overlay` methods, passing in itself as the `other_component`.

aaronayres35: So, I think what makes sense is that classes which are decedents from `AbstractOverlay` should only need to define `overlay` not `_draw_overlay`.  

It is a bit strange because `AbstractOverlay` subclasses `Component` so it does technically have a `_draw_overlay` method.  However, for `Component` `_draw_overlay` gets called via the `_draw` method.  But `_draw` is overridden by `AbstractOverlay` and simply calls `overlay` if `self.component` is not `None`.

So, in summary, I believe we can delete all the definitions of `_draw_overlay` on subclasses of `AbstractOverlay` since they are actually unused.  `overlay` is the method that actually does the work.  Since the current classes which do override `_draw_overlay` do not also override `_draw`, nothing ever actually calls `_draw_overlay`.

aaronayres35: Ah I am realizing now that we are actually using `AbstractOverlay` from chaco not enable (although it is still doing the same thing I describe above, just using `PlotComponent` instead of `Component` as a baseclass).

I do not understand currently why this is necessary / why we couldn't just use `AbstractOverlay` from enable and have overlays whose `component` trait is a chaco `PlotComponent` rather than an enable `Component`.

aaronayres35: Looking closer, some differences between `enable.AbstractOverlay` and `chaco.AbstractOverlay` include:
1) enable defines `_do_layout` (default is just `pass`), and `do_layout` methods, which chaco does not. `do_layout` (per the docstring) differs from the `Component.do_layout` method in that it allows passing an optional `component` arg.
2) in the `_draw` method, chaco calls `super` if `self.component` is `None`. enable simply returns.

Number 2 is what makes me think _maybe_ there are scenarios in which `_draw_component` gets called / is needed if we try to draw an `Overlay` for which `component` is `None`.  ie a stand alone overlay (which I will have to look for use cases / examples of ...)

aaronayres35: For instance, the `_draw_overlay` use on `Tooltip` and its subclass `DataLabel` _is_ needed, since they override `draw` and call `PlotComponent._draw` directly.

I do not see any cases in which `PlotAxis` does not have a component set (every time it is instantiated in the chaco codebase a component is passed in), however theoretically one could try drawing an Axis standalone (although i can't imagine why) in which case not having `_draw_overlay` will cause a break...

I am of two minds if it should be removed or not for the cases of `PlotAxis` and `Grid` here.  If it is to stay, because the two methods do slightly different things, I believe they will both need to stay as is with some duplication (ie `overlay` can't just call `_draw_overlay`).

aaronayres35: @rahulporuri I apologize in advance that this is an overly wordy brain dump, but what are your thoughts on this?  I hope my findings make sense

rahulporuri:
@aaronayres35 looks like there are merge conflicts on this PR. Can you address them?

chaco/axis.py
rahulporuri: I can't fully digest these findings at the moment - so my instinct is to simply defer this question to a later time. Can you write up these findings in an issue? I apologize for the delay in responding to your comments - it'll probably involve additional effort to recollect everything and put together a meaningful PR.

aaronayres35: No worries, I will open an issue

rahulporuri:
Looks like CI is happy and the merge conflicts have been resolved. Can you merge and backport this to the 5.0 maintenance branch?

PR #738, opened 2021-05-13
Add new chaco.plots submodule

aaronayres35:
This PR would close #700 

It is meant to demonstrate the plan of action for part of the proposed re-org.
If we decide to go ahead with this re-org this draft PR can be marked ready for review / fixed up and merged.

PR #744, opened 2021-05-28
Revive ScalyPlot

aaronayres35:
`ScalyPlot` was removed in #667, however it is still used / needed.

It was independent of any other aspect of `chaco.shell` and is brought back here to simply sit in `chaco.scaly_plot.py`.

~NOTE: we are planning some internal reorg (see #738), so this module may be moved in the future.~

It should be imported for the api module (I will add it to the api in this PR)

### DIFF
--- a/chaco/scaly_plot.py
+++ b/chaco/scaly_plot.py
@@ -1,0 +1,151 @@
+""" A Plot which uses ScaleSystems for its ticks.
+"""
+
+from traits.api import Any
+
+from chaco.axis import PlotAxis
+from chaco.data_range_2d import DataRange2D
+from chaco.grid import PlotGrid
+from chaco.linear_mapper import LinearMapper
+from chaco.log_mapper import LogMapper
+from chaco.plot import Plot
+from chaco.scales.scales import DefaultScale, LogScale, ScaleSystem
+from chaco.scales_tick_generator import ScalesTickGenerator
+
+
+def add_default_axes(plot, orientation="normal", vtitle="", htitle=""):
+    """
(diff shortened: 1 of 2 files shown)


rkern:
#738 isn't relevant. `ScalyPlot` is like `Plot`, not like the `PlotRenderer`s that you are moving.

Thank you!

chaco/scaly_plot.py
```diff
@@ -0,0 +1,151 @@
...
+        if self.y_axis is None:
+            self.y_axis = PlotAxis(
+                mapper=self.y_mapper,
+                orientation="left",
+                component=self,
+                tick_generator=self.y_ticks,
+            )
+
+    def _index_scale_changed(self, old, new):
+        Plot._index_scale_changed(self, old, new)
```
rahulporuri, line 144: Not entirely sure how this works. Shouldn't this be `super()._index_scale_changed(old, new)`? Same comment below with `Plot._value_scale_changed(...)`

rahulporuri (approved):
Mostly LGTM

rahulporuri:
Nice catch. How'd you notice it?

chaco/scaly_plot.py
aaronayres35: I agree, super makes more sense / is cleaner here.  This code I believe is also untested, I simply copy pasted the old file to revive it.
It could probably use a closer look / some updates / some new tests.  I can do so in some follow up PRs

aaronayres35:
> Nice catch. How'd you notice it?

There was a slack discussion which Robert tagged me in asking to revive it.

Issue #745, opened 2021-06-01
Chaco Release 5.0.0


aaronayres35:
Target date:

Release Candidate by YYYY-MM-DD

Release by YYYY-MM-DD

Release candidate
-------------------
- [x] Verify that all relevant PRs have been merged to master.
- [x] Create a PR against master to bump version number, merge that PR (#746)
- [x] From the commit just before bumping the version, create a new branch `maint/<release version number>`
- [x] Update changelog and open PR targeting a new `maint/<release version number>` branch (#750)
- [x] Update `ci-src-requirements.txt` if needed
- [x] Check MANIFEST and requirements are still up to date. (#752)
- [x] Update version in setup.py for the prerelease, open 2 PRs against `maint/<release version number>`
  - [x] Create a new branch from `maint/<release version number>`: `git checkout maint/<release version number>`; `git pull`; `git checkout -b call-it-anything-you-like`
  - [x] Set `PRERELEASE` to "rc1" and `IS_RELEASED` to true, commit, e.g. `git commit -m "Set IS_RELEASED to true for prerelease <release version number>rc1"`
  - [x] Open and merge a PR against `maint/<release version number>` (#753, #775)
  - [x] Create a new branch from `maint/<release version number>`, Flip `IS_RELEASED` back to false, commit.
  - [x] Open and merge another PR against `maint/<release version number>` (#754, #776)
- [x] Tag (annotated!) the release candidate on the commit where IS_RELEASED is set to true, e.g. `git tag -a -m "Release candidate <release version number>rc1" <release version number>rc1 <commit-hash>`
- [x] Push the tag to GitHub
- [x] Upload to PyPI
       - `git checkout <tag>`, `git clean -ffxd`, `python setup.py sdist`, `twine check dist/<...>.tar.gz`, `twine upload dist/<...>.tar.gz`
- [x] Announcement for the release candidate

Release blockers
----------------
- [x] #741 
- [x] #738 
- [x] #535 

Open Issues / PRs that would be nice to get in (but not blockers)
----------------------------------------------------------------
- [ ] #735 
- [x] #743 
- [ ] #739 (?)
- [ ] #726 

Pre-release
---
- [x] Backport PRs that have been merged to master to the maintenance branch. Use the "need backport ..." tag if there is one (but don't rely 100% on it)
- [x] Verify that no other open issue needs to be addressed before the release.
- [x] Test against other ETS packages and other ETS-using projects
[...]

PR #746, opened 2021-06-01
Bump version for continued development

aaronayres35:
part of #745 
This PR bumps the version for continued development of 5.1.0.

I will create a branch maint/5.0 from the commit just before this on master

~EDIT: This PR may have jumped the gun too soon as we might need #743 merged before the release candidate~ EDIT again: things won't break because of #737.  #743 changes remove `_draw_component` use that is unnecessary but not breaking.


rahulporuri (approved):
LGTM

PR #751, opened 2021-06-02
Contribute examples to etsdemo

### LINKED ISSUE #535: Contribute examples to the etsdemo application
The GitHub repository contains a number of example scripts which demonstrate how chaco can be used. See https://github.com/enthought/chaco/tree/master/examples. For the upcoming release of chaco, we should try to contribute the examples to the etsdemo application.

aaronayres35:
closes #535 

This PR contributes the relevant examples to the etsdemo application and does so by moving them inside the package.  I am still in the process of going through the examples and trying to decide what is relevant / useful / functioning correctly. 

I also think we may just want to leave the left over examples where they are and maybe rename the folder to legacy_examples or something like that. IIRC we did something like this with envisage. Or if a reviewer prefers they can be deleted.


aaronayres35:
Currently the documentation does not use literal includes, but the files in `examples/tutorials` / `examples/user_guide` do align with the code examples from the docs.  Perhaps those should be kept / moved / updated as needed and literal includes used in the docs?

Also once we have a final decision on the examples to include, I will delete those left behind. I have gone through the examples and I think I grabbed the majority that we will want to keep. It is possible some of those left should be kept as well. For example either `multiaxis.py` or `multiaxis_using_Plot.py`.

chaco/examples/demo/hyetograph.py
```diff
@@ -144,6 +144,8 @@ def start(self):
     )
 
 
+demo = Hyetograph()
+
+
 if __name__ == "__main__":
-    hyetograph = Hyetograph()
-    hyetograph.start()
+    demo.start()
```
aaronayres35, line 151: This is not working correctly (the demo works different when run in etsdemo vs standalone).  I imagine it is because etsdemo looks for `demo` and expects to call configure_traits or something, whereas here we really need to call start / we need to call `_perform_calculations` first...

aaronayres35:
`qt_example.py` is a strange case as etsdemo is intended to work with `HasTraits` classes specifically, rather than qt itself.  This I was unable to get to work with etsdemo, but I feel the example is still worth keeping.  I think I will add to the file docstring saying "This demo is intended to be run standalone and will not work within the etsdemo application" or something along those lines.

chaco/examples/demo/hyetograph.py
aaronayres35: I updated the example to avoid this problem.  Note this example is referenced in the documentation, so perhaps the docs themselves should be changed to match?

chaco/examples/demo/advanced/scalar_image_function_inspector.py
```diff
@@ -621,12 +621,8 @@ def configure_traits(self, *args, **kws):
...
-    model = Model(**kwargs)
-    view = PlotUI(**kwargs)
-    modelview = ModelView(model=model, view=view)
-    modelview.configure_traits()
+demo = ModelView(model=Model(), view=PlotUI())
 
 
 if __name__ == "__main__":
-    show_plot()
+    demo.configure_traits()
```
aaronayres35, line 628: This would look / work a lot better as a popup, but that leads to very strange errors due to etsdemo internals.  Doing so I get the following error:
```
  File "/Users/aayres/.edm/envs/test-chaco/lib/python3.6/site-packages/traitsui/qt4/instance_editor.py", line 428, in edit_instance
    view, kind=factory.kind, id=factory.id
  File "<string>", line 603, in edit_traits
  File "<string>", line 599, in _start_timer
  File "/Users/aayres/.edm/envs/test-chaco/lib/python3.6/site-packages/traits/trait_types.py", line 3360, in validate
    self.resolve_class(object, name, value)
  File "/Users/aayres/.edm/envs/test-chaco/lib/python3.6/site-packages/traits/trait_types.py", line 3462, in resolve_class
    super().resolve_class(object, name, value)
  File "/Users/aayres/.edm/envs/test-chaco/lib/python3.6/site-packages/traits/trait_types.py", line 3178, in resolve_class
    self.validate_failed(object, name, value)
  File "/Users/aayres/.edm/envs/test-chaco/lib/python3.6/site-packages/traits/trait_types.py", line 3207, in validate_failed
    self.error(object, name, value)
  File "/Users/aayres/.edm/envs/test-chaco/lib/python3.6/site-packages/traits/base_trait_handler.py", line 75, in error
    object, name, self.full_info(object, name, value), value
traits.trait_errors.TraitError: The 'model_view' trait of a TimerController instance must be a ModelView or None, but a value of <___main___.ModelView object at 0x7fa7c4a40150> <class '___main___.ModelView'> was specified.
Abort trap: 6
```

we have a trait defined as `a = Instance("B")` where `B` is a class defined elsewhere in the module.  However, due to etsdemo messing with locals to change `__name__` to `___name___`, we end up passing an instance of `___name___.B` which traits catches as an invalid value...
I do not know how to resolve this. Perhaps we just leave it as demo embedded in the etsdemo ui for now (it is resizable anyway so if you stretch the etsdemo ui out enough you can see the full demo).

@rahulporuri any ideas on how we could resolve this?  It is awkward because you need to do `Instance("ModelView")` not `Instance(ModelView)` because `ModelView` is defined after `TimerController` in the file, _and_ `ModelView` also has a trait which is `Instance(TimerController)` so we can't simply move the class to be defined above.

aaronayres35:
Some other demos we may want to have be `popup` instead of `demo` as they are a bit too big to fit in the etsdemo applications window.  I will go through and audit these and update as needed.  Possibly in this PR or I can do it in a quick follow up PR.

aaronayres35:
Also akin to https://github.com/enthought/enable/pull/805 we should look at moving or deleting https://github.com/enthought/chaco/blob/master/chaco/example_support.py

However, currently only `enable.example_support` is used in the chaco codebase, and as this PR currently stands only 2 examples currently being kept use it.  We probably should not be importing this from enable in the first place, but internal to ets maybe that sin can be overlooked. 
In any case we should either delete the unused chaco module, or change the uses of `enable.example_support`  inside chaco to use `chaco.example_support` instead.  Also any remaining `example_support` modules should be moved into their packages `examples` folder, eg what is done in enable PR 805 linked above

chaco/examples/demo/image_plot_origin_and_orientation.py
```diff
@@ -3,10 +3,11 @@
...
 of the plot window. These positions has the following behavior:
-    * 'left' : index increases left to right
-    * 'right' : index increases right to left
-    * 'top' : index increases top to bottom
-    * 'bottom' : index increases bottom to top
+
+* 'left' : index increases left to right
+* 'right' : index increases right to left
+* 'top' : index increases top to bottom
+* 'bottom' : index increases bottom to top
```
aaronayres35, line 10: This was previously giving an indentation warning inside the etsdemo app

chaco/examples/tests/test_etsdemo_info.py
```diff
@@ -0,0 +1,21 @@
+# (C) Copyright 2011-2020 Enthought, Inc., Austin, TX
```
rahulporuri, line 1:
```suggestion
# (C) Copyright 2005-2020 Enthought, Inc., Austin, TX
```

examples/demo/demo.py
```diff
@@ -1,23 +0,0 @@
-# -------------------------------------------------------------------------------
```
rahulporuri, line 1: I guess we deleted this file to remove any confusion about how to interact with the examples - the examples are meant to be used via the etsdemo application. Right?

rahulporuri (commented):
Image from file is broken with the following error - 

`FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Users\\rporuri\\.edm\\envs\\chaco-test-3.6-pyside2\\Scripts\\capitol.jpg'`

scatter_inspector2 fails with - 

`NameError: name '_create_plot_component' is not defined`

qt_example doesn't show anything - no errors but no visible output in the demo pane either

chaco/examples/demo/advanced/scalar_image_function_inspector.py
rahulporuri: 0. if possible, can we rename it to something other than `ModelView` because that slightly confuses things
1. I don't think it's bad to use `Instance("ModelView")` instead of `Instance(ModelView)`. I think there are definitely valid usecases (like here) where a class is defined later in the module which needs to be used here.

examples/demo/demo.py
aaronayres35: yep, that is correct

chaco/examples/demo/advanced/scalar_image_function_inspector.py
aaronayres35: Note:
 switching to 
 ```
 model_view = Instance(
        "chaco.examples.demo.advanced.scalar_image_function_inspector.DemoModelView"  # noqa: E501
    )
```
did not solve the issue.  I will leave it as a demo not a pop up for now

File: chaco/examples/demo/advanced/data_cube.py [hunk 6]
```diff
@@ -192,7 +192,7 @@ def normal_mouse_wheel(self, event):
             self.wheel_cb(self, event.mouse_wheel)
```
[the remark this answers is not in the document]
aaronayres35: I've opened #759 to keep track of the todo list

aaronayres35:
> Image from file is broken with the following error -
> 
> `FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Users\\rporuri\\.edm\\envs\\chaco-test-3.6-pyside2\\Scripts\\capitol.jpg'`

I realize looking at it now that Mark would frown at this example as it does: 
`from traits.util.resource import find_resource`
Traits should not be a place to get general utility functions, and perhaps this example should be rewritten to avoid this?  I think I will remove it from this PR entirely, and we can come back and rework it / add it back in the future.  What do you think?

aaronayres35:
@rahulporuri 
A couple of questions:
Any thoughts on https://github.com/enthought/chaco/pull/751#issuecomment-854225644 and https://github.com/enthought/chaco/pull/751#issuecomment-854060964?
Also are there any other examples you think should be moved in and kept?  If not I can go ahead and delete those which were not chosen in a final commit before this PR is merged.

rahulporuri:
> Also akin to enthought/enable#805 we should look at moving or deleting https://github.com/enthought/chaco/blob/master/chaco/example_support.py

Let's defer that for this release. We already have too much on our plate and dealing with this properly might push things by a day or two.

> It is possible some of those left should be kept as well

I want to defer this decision to after the release as well - im sure there are more examples that we can move into the package and maybe some which we can delete outright. Either way, this will need us to sit together for an hour or so after having reviewed all of the examples, which is time we are currently short on i think.

> Perhaps those should be kept / moved / updated as needed and literal includes used in the docs?

It'll definitely be a good idea to use literalinclude in the docs instead of duplicating the code. Again, one more thing to defer.

Can you create issues for all of these undecided/incomplete work items so that we can follow up later?

rahulporuri (approved):
LGTM

PR #752, opened 2021-06-03
Update requirements to be more specific

aaronayres35:
[targeting maint/5.0]
This PR updates `__requires__` to be more specific. 
We use `Property(observe=...)` and so depend on traits >= 6.2.  Also recent changes are such that we require enable 5.2 (only rc currently available).  Also enable 5.2 (and actually I think 5.1) require pyface >=7.2 as it uses `pyface.undo`.  So, chaco implicitly requires pyface >= 7.2
TBH I am not exactly sure on what traitsui is necessary (we didnt specify for enable either)

This PR will eventually need to be backported into master along with the recent changelog updates

### DIFF
--- a/chaco/__init__.py
+++ b/chaco/__init__.py
@@ -9,5 +9,5 @@ except ImportError:
     __version__ = "not-built"
 
 __requires__ = [
-    "traits>=6.2.0", "traitsui", "pyface>=7.2.0", "numpy", "enable>=5.2.0"
+    "traits>=6.2.0", "traitsui", "pyface>=7.2.0", "numpy", "enable"
 ]
--- a/chaco/__init__.py
+++ b/chaco/__init__.py
@@ -8,4 +8,6 @@ try:
 except ImportError:
     __version__ = "not-built"
 
-__requires__ = ["traits", "traitsui", "pyface", "numpy", "enable"]
+__requires__ = [
+    "traits>=6.2.0", "traitsui", "pyface>=7.2.0", "numpy", "enable>=5.2.0"
+]


chaco/__init__.py
```diff
@@ -8,4 +8,6 @@
 except ImportError:
     __version__ = "not-built"
 
-__requires__ = ["traits", "traitsui", "pyface", "numpy", "enable"]
+__requires__ = [
+    "traits>=6.2.0", "traitsui", "pyface>=7.2.0", "numpy", "enable>=5.2.0"
```
aaronayres35, line 12: I should check that a release candidate actually counts for >= .  Otherwise even with the enable 5.2.0rc2 out, there will not be available packages on PyPI satisfying the requirements when we make the rc for chaco.

If that is the case, I will unpin this requirement until after the chaco rc is made, and after enable 5.2.0 is fully released, but before chaco 5.0.0 is release

rahulporuri:
>  Also recent changes are such that we require enable 5.2 (only rc currently available). 

I'm not a 100% sure why enable 5.2 is *required*. This will only be true after #674 IIUC. Right?

rahulporuri (approved):
Almost LGTM.

I haven't tested this locally but what happens if you just pip install chaco from source in the maintenance branch now? A 5.2.0 doesn't exist in pypi yet and i'm wondering if the installation fails.

rahulporuri:
> I haven't tested this locally but what happens if you just pip install chaco from source in the maintenance branch now? A 5.2.0 doesn't exist in pypi yet and i'm wondering if the installation fails.

This doesn't work as I expected. I see the following error when I try to install a chaco sdist in a clean environment - 

```
(test-chaco) >python -m pip install chaco-4.8.1.dev200.tar.gz
Processing c:\users\rporuri\work\github\ets\chaco\dist\chaco-4.8.1.dev200.tar.gz
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... done
ERROR: Could not find a version that satisfies the requirement enable>=5.2.0 (from chaco) (from versions: 4.4.0, 4.4.1, 4.5.0, 4.5.1, 4.6.0, 4.6.1, 4.6.2, 4.7.0, 4.7.1, 4.7.2, 4.8.0, 4.8.1, 5.0.0rc2, 5.0.1, 5.1.0rc1, 5.1.0rc2, 5.1.0, 5.1.1, 5.2.0rc2)
ERROR: No matching distribution found for enable>=5.2.0
```

aaronayres35:
> > I haven't tested this locally but what happens if you just pip install chaco from source in the maintenance branch now? A 5.2.0 doesn't exist in pypi yet and i'm wondering if the installation fails.
> 
> This doesn't work as I expected. I see the following error when I try to install a chaco sdist in a clean environment -
> 
> ```
> (test-chaco) >python -m pip install chaco-4.8.1.dev200.tar.gz
> Processing c:\users\rporuri\work\github\ets\chaco\dist\chaco-4.8.1.dev200.tar.gz
>   Installing build dependencies ... done
>   Getting requirements to build wheel ... done
>     Preparing wheel metadata ... done
> ERROR: Could not find a version that satisfies the requirement enable>=5.2.0 (from chaco) (from versions: 4.4.0, 4.4.1, 4.5.0, 4.5.1, 4.6.0, 4.6.1, 4.6.2, 4.7.0, 4.7.1, 4.7.2, 4.8.0, 4.8.1, 5.0.0rc2, 5.0.1, 5.1.0rc1, 5.1.0rc2, 5.1.0, 5.1.1, 5.2.0rc2)
> ERROR: No matching distribution found for enable>=5.2.0
> ```

Yeah I just caught that myself, this is what I was unsure about in the PR description.  I will unpin this now, but we may want to repin it later (post chaco rc and enable full release)

rahulporuri (approved):
LGTM. We'll need to remember to change this once enable 5.2.0 is released 😅

aaronayres35:
> I'm not a 100% sure why enable 5.2 is _required_. This will only be true after #674 IIUC. Right?

Looking closer I believe that is correct, I don't know why I had it in my head there was something else.  I was jumping the gun too early with this PR

PR #754, opened 2021-06-03
flip IS_RELEASED back to false for continued development

aaronayres35:
This PR simply flips `IS_RELEASED` back to False following the 5.0.0 release candidate, for continued development in advance of the full 5.0.0 release

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,7 +12,7 @@ from Cython.Build import cythonize
 MAJOR = 5
 MINOR = 0
 MICRO = 0
-PRERELEASE = "rc1"
+PRERELEASE = "rc2"
 IS_RELEASED = False
 
 # If this file is part of a Git export (for example created with "git archive",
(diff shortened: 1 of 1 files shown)


rahulporuri (approved):
LGTM

setup.py
```diff
@@ -13,7 +13,7 @@
 MINOR = 0
 MICRO = 0
 PRERELEASE = "rc1"
```
rahulporuri, line 15: I wonder if we should bump the prerelease version too because rc1 has already been released but installing from this branch would make it look like rc1 hasn't been released.

```suggestion
PRERELEASE = "rc2"
```

aaronayres35: Yeah that makes more sense to me 🤔 
I have just done it this way cause it is what I have always done lol (probably copied from somewhere else the first time).  I will commit your suggestion and start doing that going forward

PR #757, opened 2021-06-07
Use stack layout helpers from enable

rahulporuri:
There was duplicate code sitting in `chaco` that handled getting the preferred stacked container size and that handled doing the layout. This PR updates `chaco` to make use of the existing helper functions in `enable`. This work has been extracted from #674 .

Ref https://github.com/enthought/enable/blob/6e1c93e9df2a6e37eb79b92a81696ac757392dce/enable/stacked_layout.py

Note that we would like this PR to be backported to the `maint/5.0` branch so that it can be included in the chaco 5.0 release.


chaco/plot_containers.py
```diff
@@ -42,6 +42,7 @@
     simple_container_get_preferred_size,
     simple_container_do_layout,
 )
+from enable.stacked_layout import stack_layout, stacked_preferred_size
```
rahulporuri, line 45: we can use `enable.api` once https://github.com/enthought/enable/pull/840 is merged.

aaronayres35 (approved):
LGTM

I will try to keep the need for these helpers in chaco in mind when reviewing the upcoming refactoring steps

rahulporuri:
> I will try to keep the need for these helpers in chaco in mind when reviewing the upcoming refactoring steps

In the next refactoring step, we will actually be able to get rid of these helpers entirely from `chaco` - because we'll directly inherit from the enable base classes which use the helpers internally.

This PR reduced the eventual diff, which i hoped would make reviews easier. See #674 for what the h/v stacked containers will eventually look like - more or less.

PR #758, opened 2021-06-07
Refactor chaco OverlayPlotContainer to inherit from enable OverlayContainer

rahulporuri:
This PR refactors the chaco `OverlayPlotContainer` to inherit from the enable `OverlayContainer`. This PR makes small changes in each commit that slowly move `OverlayPlotContainer` away from the chaco `BasePlotContainer` by redefining the traits on the latter on the former. Finally, we update the inheritance and remove redundant methods on the `OverlayPlotContainer`.

Note that we would like this PR to be backported to the `maint/5.0` branch so that it can be included in the final `maint/5.0` release.

### DIFF
--- a/chaco/plot_containers.py
+++ b/chaco/plot_containers.py
@@ -33,6 +33,7 @@ from traits.api import (
     Instance,
     List,
     Property,
+    Str,
     String,
     Trait,
     Tuple,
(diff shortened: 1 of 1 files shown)


aaronayres35 (approved):
LGTM

PR #764, opened 2021-06-07
Make chaco v/h plot containers inherit from enable v/h stacked containers

rahulporuri:
This PR refactors the `HPlotContainer` and `VPlotContainer` classes in enable to inherit from the enable `HStackedContainer` and `VStackedContainer` respectively. This removes duplicate code between chaco and enable. In the end, in our opinion, the `*PlotContainer` classes in chaco are more-readable than they were before.

Note that this PR was written with enable main branch in mind - but it should work with the current release of `enable` as well. This PR is what will introduce the dependency on `enable >= 5.2` on the upcoming `chaco` major release.

This PR can be reviewed one commit at a time as we have taken great care to keep the commits self-contained. Each commit makes only one type of change, which we hope will make the review process simpler.

This is the last big PR extracted from #674.

Note that we would like this PR to be backported to the `maint/5.0` branch so that it can be included in the chaco 5.0.0 major release.

### DIFF
--- a/chaco/plot_containers.py
+++ b/chaco/plot_containers.py
@@ -121,6 +121,12 @@ class HPlotContainer(BasePlotContainer):
     # attribute.
     stack_index = 0
 
+    #: Redefine the container layers to name the main layer as "plot" instead
+    #: of the Enable default of "mainlayer"
+    container_under_layers = Tuple("background", "image", "underlay", "plot")
+
+    draw_layer = Str("plot")
+
     draw_order = Instance(list, args=(DEFAULT_DRAWING_ORDER,))
 
     # The dimension along which to stack components that are added to
@@ -182,6 +188,12 @@ class VPlotContainer(BasePlotContainer):
     """
     A plot container that stacks plot components vertically.
     """
+
+    #: Redefine the container layers to name the main layer as "plot" instead
+    #: of the Enable default of "mainlayer"
+    container_under_layers = Tuple("background", "image", "underlay", "plot")
+
+    draw_layer = Str("plot")
 
     draw_order = Instance(list, args=(DEFAULT_DRAWING_ORDER,))
 
--- a/chaco/plot_containers.py
+++ b/chaco/plot_containers.py
@@ -163,6 +163,13 @@ class HPlotContainer(StackedPlotContainer):
 
     draw_order = Instance(list, args=(DEFAULT_DRAWING_ORDER,))
 
+    # The dimension along which to stack components that are added to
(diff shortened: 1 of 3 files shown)


rahulporuri:
CI is failing with the following error - 

```
FAIL: test_valign (chaco.tests.test_plotcontainer.HPlotContainerTestCase)
  File "/home/runner/.edm/envs/chaco-test-3.6-null/lib/python3.6/site-packages/chaco/tests/test_plotcontainer.py", line 175, in test_valign
    self.assertEqual(comp1.position, [0, 50])
AssertionError: [0, 0] != [0, 50]

```

This failure doesn't exist if we test against enable from source on the main branch. Like I mentioned earlier, this PR is what introduces the dependency on `enable >= 5.2.0`.

chaco/plot_containers.py
```diff
@@ -129,51 +124,8 @@ class HPlotContainer(Container):
 
     draw_order = Instance(list, args=(DEFAULT_DRAWING_ORDER,))
 
-    # The dimension along which to stack components that are added to
-    # this container.
-    stack_dimension = Str("h", transient=True)
```
aaronayres35: we will want to set `transient=True` in `enable.stacked_container.StackedContainer`

aaronayres35 (approved):
LGTM, thank you very much for the commit breakdown! It made the changes far more tractable 

Is the plan then to hold off on merging until enable 5.2 is available on edm, and then revert the `edmtool` / gh actions changes before pushing this through? We can also update enable to use `transient` metadata effectively prior to 5.2.0 release and remove the `__getstate__` methods in this PR before it is merged. 

Edit: I've opened https://github.com/enthought/enable/pull/841

rahulporuri:
> Is the plan then to hold off on merging until enable 5.2 is available on edm, and then revert the edmtool / gh actions changes before pushing this through?

I'm actually hoping that we can merge this PR as-is - and eventually update ci to use the enable 5.2.0 release tag - and after the chaco release is also complete, revert everything to use edm eggs instead of installing from git source.

I don't see any other way by which we can make a chaco rc2 that includes these refactor PRs.

> We can also update enable to use `transient` metadata effectively / pull up the `__getstate__` prior to 5.2.0 release and remove the `__getstate__` methods in this PR before it is merged.

Yup. These would be nice and thanks for opening the PR. They can be removed after the PR is merged. I don't see a reason why they should block the release.

Are you okay with me merging this PR as-is and cherry-picking the commits back into the 5.2.0 maintenance branch?

PR #770, opened 2021-06-09
Bring back marker_trait to chaco.api

### LINKED ISSUE #771: Downstream users still import marker_trait from chaco.api
This was removed in PR #679

To avoid breaking code we should bring it back into the `chaco.api` for the 5.0.0 release

aaronayres35:
[targeting `maint/5.0`]

closes #771 

There are still users trying to import `marker_trait` from chaco not enable.  To avoid causing breakages with the upcoming release, we are undoing this change for now.

### DIFF
--- a/chaco/api.py
+++ b/chaco/api.py
@@ -527,3 +527,7 @@ from .default_colormaps import (
     Set3,
 )
 from .default_colors import cbrewer, palette11, palette14, PALETTES
+
+# Importing into the Chaco namespace for backwards compatibility.  New code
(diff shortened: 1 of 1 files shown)


rahulporuri:
Same comment here as with the enable PR bringing back `kiva.agg.points_in_polygon` - here's the search https://github.com/search?p=1&q=org%3Aenthought+marker_trait&type=Code and I only see a few internal uses that I see relevant. Given this, I think it'd be better if we can fix the usage internally instead of having to bring these back.

PR #772, opened 2021-06-09
Fix _draw_overlay in chaco.axis

aaronayres35:
This got broken in #743 

`component` was always None in this case (hence the removed `if component is not None:`.  I had just missed this before.

### DIFF
--- a/chaco/axis.py
+++ b/chaco/axis.py
@@ -292,7 +292,7 @@ class PlotAxis(AbstractOverlay):
 
         if not self._cache_valid:
             self._calculate_geometry()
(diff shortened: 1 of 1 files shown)


rahulporuri (approved):
LGTM.

rahulporuri:
Just confirming that this will be backported to `maint/5.0` to be included in chaco 5.0.0.

aaronayres35:
> Just confirming that this will be backported to `maint/5.0` to be included in chaco 5.0.0.

This PR is actually targeting `maint/5.0` but this will need to be backported to master

PR #776, opened 2021-06-10
flip IS_RELEASED back to false for continued development

aaronayres35:
This PR simply flips `IS_RELEASED` back to `False` following the 5.0.0rc2, for continued development in advance of the full 5.0.0 release

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,8 +20,8 @@ from Cython.Build import cythonize
 MAJOR = 5
 MINOR = 0
 MICRO = 0
-PRERELEASE = "rc2"
-IS_RELEASED = True
+PRERELEASE = ""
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -20,8 +20,8 @@
 MAJOR = 5
 MINOR = 0
 MICRO = 0
-PRERELEASE = "rc2"
-IS_RELEASED = True
+PRERELEASE = ""
```
rahulporuri, line 23: let's continue bumping up - even if we know that the next release will be the final release.

```suggestion
PRERELEASE = "rc3"
```

rahulporuri (approved):
LGTM

PR #784, opened 2021-06-14
Remove u prefixes as they are not needed on python 3

### LINKED ISSUE #613: Remove unnecessary u-prefixes from the codebase and examples
The u-prefix e.g. `u"abc"` is not needed on Python 3 so the prefixes can be removed. Note that this is not worth doing unless this can be automated using a search-and-replace, possibly using a regex.

aaronayres35:
closes #613


rahulporuri (approved):
LGTM

Issue #785, opened 2021-06-15
A large number of errors observed in the testsuite with the celiagg backend.


rahulporuri:
A number of errors were observed in the testsuite when `ETS_TOOLKIT=qt.celiagg`, Note that these failures were observed on windows.

```
======================================================================
ERROR: test_text_plot (chaco.tests.test_plot.PlotTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Users\rporuri\.edm\envs\chaco-test\lib\site-packages\chaco\tests\test_plot.py", line 117, in test_text_plot
    gc = PlotGraphicsContext((250, 250))
  File "C:\Users\rporuri\.edm\envs\chaco-test\lib\site-packages\chaco\plot_graphics_context.py", line 42, in __init__
    size_or_ary, *args, **kw
  File "C:\Users\rporuri\.edm\envs\chaco-test\lib\site-packages\kiva\celiagg.py", line 93, in __init__
    buffer = np.zeros(shape, dtype=np.uint8)
TypeError: 'float' object cannot be interpreted as an integer

----------------------------------------------------------------------
```

The culprit seems to be `size_or_ary` in the `PlotGraphicsContextMixin` object.

https://github.com/enthought/chaco/blob/d3b7c374eaeb2b6a4ba194eed5751763f3d9b8d2/chaco/plot_graphics_context.py#L35-L43

rahulporuri:
Note that forcing `scale` to `int` makes the testsuite fail with only one error (instead of the 77 like before)

https://github.com/enthought/chaco/blob/d3b7c374eaeb2b6a4ba194eed5751763f3d9b8d2/chaco/plot_graphics_context.py#L34

```
======================================================================
FAIL: test_draw_border_simple (chaco.tests.test_border.DrawBorderTestCase)
Borders should have the correct height and width.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Users\rporuri\.edm\envs\chaco-test\lib\site-packages\chaco\tests\test_border.py", line 52, in test_draw_border_simple
    self.assertRavelEqual(actual, desired)
  File "C:\Users\rporuri\.edm\envs\chaco-test\lib\site-packages\chaco\tests\test_border.py", line 29, in assertRavelEqual
    alltrue(ravel(x) == ravel(y)), "\n%s\n !=\n%s" % (x, y)
AssertionError: False is not true :
[[  0   0   0   0   0   0]
 [  0   0   0   0   0   0]
[...]

PR #788, opened 2021-06-15
Specify size_or_ary contains ints

aaronayres35:
part of #785 
As mentioned in the issue, we can resolve many test suite errors with celiagg by enforcing these be ints.  Note that scale itself I think can be a float (e.g. the `scale_ctm` method expects floats as arguments), however, `size_or_ary` gets used as a `_width` and `_height` and is assumed to contain integers. 

~Note, with this change I now see 2 failures in the test suite:~ EDIT: I only see the one error mentioned on the issue:
```
FAIL: test_draw_border_simple (chaco.tests.test_border.DrawBorderTestCase)
Borders should have the correct height and width.
  File "/Users/aayres/Desktop/chaco/chaco/tests/test_border.py", line 52, in test_draw_border_simple
    self.assertRavelEqual(actual, desired)
  File "/Users/aayres/Desktop/chaco/chaco/tests/test_border.py", line 29, in assertRavelEqual
    alltrue(ravel(x) == ravel(y)), "\n%s\n !=\n%s" % (x, y)
AssertionError: False is not true : 
[[  0   0   0   0   0   0]
 [  0   0   0   0   0   0]
 [  0   0 255 255   0   0]
 [  0   0 255 255   0   0]
 [  0   0   0   0   0   0]
 [  0   0   0   0   0   0]]
 !=
[[255 255 255 255 255 255]
 [255   0   0   0   0 255]
 [255   0 255 255   0 255]
 [255   0 255 255   0 255]
 [255   0   0   0   0 255]
 [255 255 255 255 255 255]]
```

### DIFF
--- a/chaco/plot_graphics_context.py
+++ b/chaco/plot_graphics_context.py
@@ -34,8 +34,8 @@ class PlotGraphicsContextMixin(object):
         scale = kw.pop("dpi", 72.0) / 72.0
         if type(size_or_ary) in (list, tuple) and len(size_or_ary) == 2:
             size_or_ary = (
-                size_or_ary[0] * scale + 1,
-                size_or_ary[1] * scale + 1,
+                int(size_or_ary[0] * scale + 1),
+                int(size_or_ary[1] * scale + 1),
             )
 
         super().__init__(


rahulporuri (approved):
LGTM. I verified that kiva `GraphicsContext` expects heigth, width of image to be ints, not floats.

chaco/plot_graphics_context.py
```diff
@@ -34,8 +34,8 @@ def __init__(self, size_or_ary, *args, **kw):
         scale = kw.pop("dpi", 72.0) / 72.0
```
rahulporuri, line 34: If we can assume that `size_or_ary` is a tuple of **ints**, we can make the following change instead (IIUC the code correctly) but we can be defensive and go with the changes you propose.

```suggestion
        scale = kw.pop("dpi", 72.0) // 72.0
```

aaronayres35: If it's alright with you I think I will leave the current approach, as you mention, to be defensive.  The reasoning being that I think `scale_ctm` can accept floats and `size_or_ary` is what we specifically want to have integer entries.  
I would guess in practice `scale` will typically be an integer but I am not certain that would always be the case

PR #789, opened 2021-06-15
Explicitly require enable 5.2.0 which is now on PyPI

aaronayres35:
Now that enable 5.2.0 has been uploaded to pypi, we can explicitly add the requirement. 

This PR should be pulled back into master as well

### DIFF
--- a/chaco/__init__.py
+++ b/chaco/__init__.py
@@ -10,7 +10,7 @@ except ImportError:
 
 
 __requires__ = [
-    "traits>=6.2.0", "traitsui", "pyface>=7.2.0", "numpy", "enable"
+    "traits>=6.2.0", "traitsui", "pyface>=7.2.0", "numpy", "enable>=5.2.0"
 ]
 
 


rahulporuri (approved):
LGTM. Note that we'll also need the changes to `ci\edmtool.py` along with this to ensure that CI also tests against `enable` from PyPI.

aaronayres35:
> LGTM. Note that we'll also need the changes to `ci\edmtool.py` along with this to ensure that CI also tests against `enable` from PyPI.

Should those changes be included in the release or just on master?

rahulporuri:
>  Since this is a CI specific fix I am going to go ahead with the release and make that update to master soon after. Is that okay with you @rahulporuri or would you prefer this to be included in the release?

This doesn't have to be included in the release and the changes themselves will be simplified once the edm eggs for enable 5.2.0 become available - but the changes need to be made both on the master and the maint/5.0 branches.

PR #793, opened 2021-06-15
Fix a couple more links in documentation

aaronayres35:
Ref: #792 

This PR fixes a couple other links to examples in the documentation.  Note these changes _were_ included in the gh-pages PR

PR #792, opened 2021-06-15
Update Documentation for 5.0.0

aaronayres35:
This PR updates the documentation for the 5.0.0 release.  Unfortunately, I realized a couple additional broken links in the built docs.  I did the unorthodox thing of fixing them before building the docs for this PR.  After this, I will open a PR to fix them in the docs source in the code base.  I figured the release is already tagged, but we want the served docs to not have the broken links.  I imagine the scenario of someone building the docs themselves not from master or the maint branch will be a rare occurrence.  

The steps I took were:
```
git clean -ffxd
git checkout 5.0.0
python ci/edmtool.py install
edm shell -e chaco-test-3.6-null
edm install scipy pyqt5
python setup.py install
exit
# fix a couple more broken links locally.  See PR #793 
python ci/edmtool.py docs
# move docs/build/html folder to my desktop
git stash # stash my fixes, this was used later for PR #793
git checkout gh-pages
git clean -ffxd
git pull
git checkout -b gh-pages-5.0
mkdir 4.8
mkdir 5.0
mv `ls -d [![:digit:]]*` 4.8/. # move old 4.8 built 
git add -u
git add 4.8/
git commit -m "move 4.8 build docs into a 4.8 folder"
# move html folder containing built docs into cwd
rsync -r --exclude="\.buildinfo" html/* 5.0/  # put 5.0 built docs in 5.0 folder
rm -r html
cp -r 5.0/* .
git add .
git commit -m "add 5.0 built docs to 5.0 folder and root directory"
```

To test:
`python -m http.server`

PR #794, opened 2021-06-15
fix more broken links

aaronayres35:
Follow up to #793 
These changes also went into #792.

Only some of the examples included in the documentation were moved into the package.  I went through and tested the links on the annotated examples page and now they finally all work as expected.  Additionally the tools page link pointed out on 792/ 793 is fixed here.


rahulporuri (approved):
LGTM

PR #795, opened 2021-06-17
Use enable from edm instead of git/github source

rahulporuri:
This PR removes the temporary changes introduced in https://github.com/enthought/chaco/pull/764


aaronayres35 (approved):
LGTM

PR #801, opened 2021-06-28
Add an ets-from-source cron job on GH Actions

aaronayres35:
This PR adds a cron job on GH Actions (ultimately we would like to drop travis).

### DIFF
--- a/.github/workflows/ets-from-source.yml
+++ b/.github/workflows/ets-from-source.yml
@@ -1,0 +1,83 @@
+name: Test with ETS packages from source
+
+on:
+  pull_request:
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+  schedule:
+    # Every Friday at 00:00 UTC
+    - cron:  '0 0 * * 5'
+
+env:
+  INSTALL_EDM_VERSION: 3.2.3
+
+jobs:
+
+  # Test against EDM packages from source on Linux
+  test-with-edm:
+    strategy:
+      matrix:
+        os: ubuntu-latest
+        toolkit: ['null', 'pyqt5', 'pyside2']
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install Qt dependencies for Linux
+        run: |
+          sudo apt-get update
+          sudo apt-get install qt5-default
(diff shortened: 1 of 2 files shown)


.github/workflows/ets-from-source.yml
```diff
@@ -0,0 +1,83 @@
+name: Test with ETS packages from source
+
+on:
+  pull_request:
```
rahulporuri, line 4: This will be removed before the PR is merged IIUC. Right?

.github/workflows/ets-from-source.yml
```diff
@@ -0,0 +1,83 @@
...
+    # Every Friday at 00:00 UTC
+    - cron:  '0 0 * * 5'
+
+env:
+  INSTALL_EDM_VERSION: 3.2.3
+
+jobs:
+
+  # Test against EDM packages from source on Linux
+  test-with-edm:
```
rahulporuri, line 17: Don't we need a different name here when compared to the normal jobs? The normal jobs also use `test-with-edm` so it would be better if we used different names here.

.github/workflows/ets-from-source.yml
```diff
@@ -0,0 +1,83 @@
...
+    strategy:
+      matrix:
+        os: ubuntu-latest
```
rahulporuri, line 20: Looks like the workflow wasn't actually running because of this error. See https://github.com/enthought/chaco/actions/runs/979545563

```suggestion
        os: [ubuntu-latest]
```

rahulporuri (changes requested):
The job is failing with the following error - 

```
The workflow is not valid.
.github/workflows/ets-from-source.yml (Line: 20, Col: 13): Unexpected value 'ubuntu-latest'
.github/workflows/ets-from-source.yml (Line: 61, Col: 3): 'test-with-edm' is already defined
```

Ref https://github.com/enthought/chaco/actions/runs/979545563

.github/workflows/ets-from-source.yml
aaronayres35: yep 👍

.github/workflows/ets-from-source.yml
aaronayres35: Ah yes, nice catch thanks!

rahulporuri (approved):
LGTM

PR #802, opened 2021-06-28
Fix BaseXYPlot map_screen

### LINKED ISSUE #272: BaseXYPlot creates invalid screen points

aaronayres35:
fixes #272 

This PR pulls out the changes specific to `BaseXYPlot` and its subclasses from #726

### DIFF
--- a/chaco/base_xy_plot.py
+++ b/chaco/base_xy_plot.py
@@ -11,7 +11,7 @@
 """ Defines the base class for XY plots.
 """
 from math import sqrt
-from numpy import around, array, isnan, transpose
+from numpy import around, array, empty, isnan, transpose
 
 # Enthought library imports
 from enable.api import black_color_trait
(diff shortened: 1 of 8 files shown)


rahulporuri (approved):
LGTM

PR #804, opened 2021-06-28
Add regression test / fix for DataLabelTool

aaronayres35:
This PR adds a regression test and a fix for the specific error reported on #550.

550 is a bit broader in scope though and would also include #802 (and probably more) to be closed.
Note the changes here were pulled from #726 in an effort to make smaller more reviewable PRs.  However, there is some interconnectedness between these PRs.  This PR should not be merged first, as it will break things without the change in #802

As the issue mentions, currently the `DataLabelTool` is forced to assume the `BasaeXYPlot` way or the `Base2DPlot` way. 
"This difference forces objects that interact with `AbstractPlotRenderers` to make an assumption about the behavior of the `map_screen` method. For example, the `DataLabelTool` assumes the map_screen return shape matches that of `BaseXYPlot` for a single point ((2,)) ... And therefore this tool throws an exception when used with a `Base2DPlot`"

This PR makes `DataLabelTool` follow the `Base2DPlot` way (which we are pushing as the new standard), and #802 makes `BaseXYPlot` do the same.  On this branch, `DataLabelTool` will not work with a `BaseXYPlot`.

EDIT: I've added a failing test for use of `DataLabelTool`with a `BaseXYPlot`.  This test will pass once #802 is merged

### DIFF
--- a/chaco/tools/data_label_tool.py
+++ b/chaco/tools/data_label_tool.py
@@ -71,7 +71,7 @@ class DataLabelTool(DragTool):
         """
         if self.component:
             label = self.component
-            pointx, pointy = label.component.map_screen(label.data_point)
+            pointx, pointy = label.component.map_screen(label.data_point)[0]
             self._original_offset = (label.x - pointx, label.y - pointy)
             event.window.set_mouse_owner(self, event.net_transform())
             event.handled = True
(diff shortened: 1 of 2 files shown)


aaronayres35:
#802 was merged, so I merged master with this branch and now CI passes as expected.

rahulporuri (approved):
LGTM

PR #805, opened 2021-06-29
Make map_screen return an empty array not empty list in empty input case

aaronayres35:
This pulls some additional changes out of #726.

This is to keep the return type for the method consistent


rahulporuri (approved):
LGTM

PR #808, opened 2021-07-07
fix hittest_tool.py example use of map_screen

aaronayres35:
This example became broken by the recent changes to map_screen.
This change was pulled from #726 
Note there are some other examples which did not get pulled into `chaco/examples` that could be fixed as well.  I did not do that as I could not recall if they were planned to be deleted entirely.  See #726 changes to `examples/tutorials/...` 
I am happy to make those changes on this PR if need be.


rahulporuri (approved):
LGTM.

Let's audit the other potentially broken examples later.

PR #810, opened 2021-07-07
Enforce data array is array in BandedMapper / PolarMapper

aaronayres35:
This PR contains the last bit of changes which need to be pulled from #726
It simply updates `BandedMapper` / `PolarMapper` to follow the lead of their base class `LinearMapper`.

After all these are merged, I will do a final sweep of the code base to see that nothing else needs to be updated with the recent return type changes (plan is to do code base searches for `map_screen`, etc. / to go through examples and ensure they all work as expected.  Test suite will pass as it has passed on every PR, but just to be sure we didn't miss / break anything in the process not covered by the test suite).

### DIFF
--- a/chaco/plots/horizon_plot.py
+++ b/chaco/plots/horizon_plot.py
@@ -29,6 +29,8 @@ class BandedMapper(LinearMapper):
             else:
                 return array([self.low_pos])
         else:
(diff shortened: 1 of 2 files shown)


rahulporuri (approved):
LGTM

PR #816, opened 2021-08-10
 Don't link to examples on the main branch in documentation

rahulporuri:
At the moment, the documentation links to examples on the `main`/default branch. This isn't right because the examples might change in between releases, potentially confusing users who are referring to the documentation. This PR fixes that problem by making the documentation use version-specific links i.e. for the 5.0.0 release, the `5.0.0` tag will be used instead of the `main` branch.

I tested this locally by checking the links generated in the built docs - the examples links will be broken as expected as the dev version has no corresponding tag on GitHub. I then hardcoded the version in the sphinx configuration to `5.0.0` and the links worked as expected, pointing to the `5.0.0` tree instead of `main`.

Note that this commit also creates the new `github-demo` external link using the `extlinks` sphinx extension. This is how we create the version-specific links in this PR and this is what we do in other ETS projects.

Note also that this PR also fixes broken links in the annotated examples page of the documentation.

### DIFF
--- a/docs/source/annotated_examples.rst
+++ b/docs/source/annotated_examples.rst
@@ -17,7 +17,7 @@ examples as starting points for their own applications.
 -----------------------
 An example showing Chaco's BarPlot class.
 
-source: `bar_plot_stacked.py <https://github.com/enthought/chaco/tree/main/examples/demo/basic/bar_plot_stacked.py>`_
+source: :github-demo:`bar_plot_stacked.py <chaco/examples/demo/basic/bar_plot_stacked.py>`.
 
 .. image:: example_images/bar_plot.png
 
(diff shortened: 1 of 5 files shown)


docs/source/annotated_examples.rst
```diff
@@ -17,7 +17,7 @@ examples as starting points for their own applications.
 -----------------------
 An example showing Chaco's BarPlot class.
 
-source: `bar_plot_stacked.py <https://github.com/enthought/chaco/tree/main/examples/demo/basic/bar_plot_stacked.py>`_
+source: :github-demo:`bar_plot_stacked.py <chaco/examples/demo/basic/bar_plot_stacked.py>`.
```
mdickinson, line 20: I haven't met the `github-demo` role before. Where is this defined? (Is there documentation somewhere?)

rahulporuri: it's a custom role that we define - https://github.com/enthought/chaco/blob/737bcc73c8222cf3a236cc6feee5103b41f88f10/docs/source/conf.py#L227-L231

mdickinson:
> the examples links will be broken as expected as the dev version has no corresponding tag on GitHub

This makes me wonder whether this is really the right thing to do.

mdickinson (approved):
This PR LGTM, but I'm wondering whether we should rethink the approach altogether: it might be better if the built documentation were self-contained and didn't rely on links to GitHub. I'm not sure what particular advantage the GitHub links are providing for documentation users.

PR #827, opened 2022-02-06
FIX: Make spectrum demo runnable

jvkersch:
The spectrum demo had two small issues (`linspace` complains about being fed floats and `scipy.fft` is now apparently a module).﻿


corranwebster (approved):
Thanks for the fixes.

LGTM

Will merge once CI completes.

PR #835, opened 2022-09-15
FIX: suppress invalid data warnings for handling NaNs in DataRange classes

### LINKED ISSUE #834: Plotting data with nans causes numpy warnings
**Problem Description**

There is a (quite secondary) method in `DataRange1D` that will cause `numpy` to emit a warning for every scatter plot creation if `nan`s are present. I'm not sure if it's a design principle explicitly claimed in writing, but Chaco seems to be built to be tolerant to `nan`s in data, so it feels like `nan`s should not cause warnings.

**Reproduction Steps:**

Running the following:

```python
from numpy import sort
from numpy.random import random
from enable.api import Component, ComponentEditor
from traitsui.api import UItem, View

class Demo(HasTraits):
    plot = Instance(Component)
    traits_view = View(
        UItem("plot", editor=ComponentEditor(size=(650, 650))),
        resizable=True)

    def _plot_default(self):
        x, y = sort(random(50)), random(50)
        x[5] = float("nan")
        y[5] = float("nan")

        pd = ArrayPlotData()
        pd.set_data("index", x)
        pd.set_data("value", y)
        plot = Plot(pd)
        plot.plot(("index", "value"), type="scatter")
        return plot

Demo().configure_traits()
```

works well, but emits:
```
/home/ndemitri/.edm/envs/ds-msrl-dev/lib/python3.6/site-packages/chaco/data_range_1d.py:126: RuntimeWarning: invalid value encountered in greater_equal
  return (data.view(ndarray) >= self._low_value) & (
/home/ndemitri/.edm/envs/ds-msrl-dev/lib/python3.6/site-packages/chaco/data_range_1d.py:127: RuntimeWarning: invalid value encountered in less_equal
  data.view(ndarray) <= self._high_value
```

(Note: the same happens when running the [nans_plot](https://github.com/enthought/chaco/blob/main/chaco/examples/demo/basic/nans_plot.py) example verbatim)

**Expected behavior:**

No warning should be emitted.

---

Note: the warning is emitted here:

https://github.com/enthought/chaco/blob/8a8162f465ce7c4b6a17eef36a2017a38c1a4201/chaco/data_range_1d.py#L126-L127

and is the result of comparing (twice) a `nan`-containing array with a float. The warning makes sense in general `numpy` usage, but the thing it is warning us against seems to be fine for DataRange1D use case: `nan`s value get a `False` mask, i.e. they are not within the plotted range.

nicolasap-dm:
Fixes #834 

- suppresses "invalid value" warnings in numpy when computing range masks in the `BaseDataRange` API implementations
   - [x] in `DataRange1D`
   - [x] in `DataRange2D` ¹ 
- doesn't change behavior, apart from warning suppression

The default behavior of float comparison with `nan`s is consistent with the `mask_data` abstract specification (being "inside the range" implies being finite), so it should be fine to confidently use it without warnings.

By the way, all uses of `mask_data` in the Chaco codebase outside of the classes themselves are already paired with a `& nan_mask`. I'm not sure if that's redundant, but it makes me confident that exluding `nan`s from masks is expected everywhere.

---

¹ ~a curious finding: apparently transposing an array defuses the warning in the first place~

<details>

```python
>>> import numpy as np
>>> nan = float("nan")

>>> np.array([1.0, nan]) > 0.5
__main__:1: RuntimeWarning: invalid value encountered in greater
array([ True, False])

>>> np.array([[1.0], [nan]]).transpose()[0] > 0.5
array([ True, False]) 
```

Edit: I was wrong, there is an issue of state. Whichever of the above comparisons gets run first will emit the warning.

</details>

### DIFF
--- a/chaco/data_range_1d.py
+++ b/chaco/data_range_1d.py
@@ -124,10 +124,10 @@ class DataRange1D(BaseDataRange):
         Implements AbstractDataRange.
         """
         with errstate(invalid="ignore"):
-            # The data array may contain NaNs. These are strictly invalid for
-            # comparison and Numpy would emit a warning. Since we are happy
-            # with the defaul behavior (NaNs are "False" in the mask), we
-            # silence the warning.
+            # Running under context because the data array may contain NaNs.
+            # These are strictly invalid for comparison and Numpy would emit
+            # a warning. Since we are happy with the default behavior (NaNs
+            # become "False" in the mask), we silence the warning.
             mask = (data.view(ndarray) >= self._low_value) & (
                 data.view(ndarray) <= self._high_value
             )
--- a/chaco/data_range_1d.py
+++ b/chaco/data_range_1d.py
@@ -16,7 +16,7 @@ Defines the DataRange1D class.
 # Major library imports
 from math import ceil, floor, log
 
-from numpy import compress, inf, isinf, isnan, ndarray, errstate
+from numpy import compress, errstate, inf, isinf, isnan, ndarray
 
 # Enthought library imports
 from traits.api import Bool, CFloat, Enum, Float, Property, Trait, Callable
--- a/chaco/data_range_1d.py
+++ b/chaco/data_range_1d.py
@@ -16,7 +16,7 @@ Defines the DataRange1D class.
 # Major library imports
 from math import ceil, floor, log
 
-from numpy import compress, inf, isinf, isnan, ndarray
+from numpy import compress, inf, isinf, isnan, ndarray, errstate
 
 # Enthought library imports
 from traits.api import Bool, CFloat, Enum, Float, Property, Trait, Callable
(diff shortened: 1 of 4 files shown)


nicolasap-dm:
The tests check that `NaN`s are `False` in masks.

They _try_ to check also that warnings are not emitted, however it is not guaranteed that they would spot a regression, because if a matching warning has already been emitted anywhere else in the test suite, the "once" rule on warnings will already have prevented it to be emitted again.

From the docs, it seems to be overcomable:

> One thing to be aware of is that if a warning has already been raised because of a `once`/`default` rule, then no matter what filters are set the warning will not be seen again unless the warnings registry related to the warning has been cleared.
>
> — https://docs.python.org/3/library/warnings.html#testing-warnings

however, clearing the registry doesn't seem to be easy or supported: https://bugs.python.org/issue21724.

CC @mdickinson since this came up in offline discussion.

---

Mark, you're welcome to review if you have time. The core of the PR is based on @rkern's suggestion [on Slack](https://enthought.slack.com/archives/C024RRPHD/p1663186988721319?thread_ts=1663177548.121569&cid=C024RRPHD).

corranwebster (approved):
Looks good!

Thanks for the contribution.

PR #839, opened 2022-10-28
Add action to publish to PyPI

corranwebster:
This will need appropriate secrets added.

### DIFF
--- a/.github/workflows/release-to-pypi.yml
+++ b/.github/workflows/release-to-pypi.yml
@@ -16,7 +16,7 @@ jobs:
     steps:
 
     - name: Check out the release commit
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
 
     - name: Set up QEMU
       uses: docker/setup-qemu-action@v1
@@ -25,14 +25,12 @@ jobs:
       if: runner.os == 'Linux'
 
     - name: Set up Python
-      uses: actions/setup-python@v2
+      uses: actions/setup-python@v4
       with:
-        python-version: '3.8'
+        python-version: '3.10'
 
     - name: Install Python packages needed for wheel build and upload
-      run: |
-        python -m pip install --upgrade pip setuptools
-        python -m pip install twine wheel
+      run: python -m pip install twine
 
     - name: Build wheels
       uses: pypa/cibuildwheel@v2.11.2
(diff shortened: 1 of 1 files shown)


.github/workflows/release-to-pypi.yml
```diff
@@ -0,0 +1,80 @@
...
+    strategy:
+      matrix:
+        os: [windows-latest, macos-latest, ubuntu-latest]
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+
+    - name: Check out the release commit
+      uses: actions/checkout@v2
```
mdickinson, line 19: Suggest using the latest version here: 
```suggestion
      uses: actions/checkout@v3
```

.github/workflows/release-to-pypi.yml
```diff
@@ -0,0 +1,80 @@
...
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v1
+      with:
+        platforms: arm64
+      if: runner.os == 'Linux'
+
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.8'
```
mdickinson, line 30: Consider using Python 3.10, for longevity?

.github/workflows/release-to-pypi.yml
```diff
@@ -0,0 +1,80 @@
...
+      uses: actions/checkout@v2
+
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v1
+      with:
+        platforms: arm64
+      if: runner.os == 'Linux'
+
+    - name: Set up Python
+      uses: actions/setup-python@v2
```
mdickinson, line 28: Suggest using latest version
```suggestion
      uses: actions/setup-python@v4
```

.github/workflows/release-to-pypi.yml
```diff
@@ -0,0 +1,80 @@
...
+      with:
+        python-version: '3.8'
+
+    - name: Install Python packages needed for wheel build and upload
+      run: |
+        python -m pip install --upgrade pip setuptools
```
mdickinson, line 34: Nitpick: This step shouldn't be needed: if we have a pyproject.toml, we don't need `setuptools`, and `pip` is automatically updated to the latest version by the setup-python action.

.github/workflows/release-to-pypi.yml
```diff
@@ -0,0 +1,80 @@
...
+        python -m pip install twine wheel
```
mdickinson, line 35: I don't think we need `wheel` here.

mdickinson (approved):
Looks fine to me in principle; it's annoying that there's no way to test _before_ merging to main.

I've left a few nitpicks on the first job; most of those also apply to the second, but I didn't leave duplicate commments.

PR #841, opened 2022-10-28
Fix PyPI release script

corranwebster:
Need numpy and cython to build sdist due to setup.py requirements.  Need to specify the format of the long description.

Also try to trim down the platforms to those that have a numpy wheel available (building numpy under emulation is _painful_).


mdickinson (approved):
LGTM

PR #847, opened 2022-12-19
MAINT: Use "bool" instead of deprecated "np.bool"

rahulporuri:
This PR replaces `np.bool` with the built in `bool`


corranwebster:
`maint/*` branches are protected, so I can't push changes to make the tests pass.  I will close this and open a new PR.

Issue #824, opened 2022-01-21
auto_ticks raises ValueError for some inputs


mdickinson:
We're running into the following problem with `auto_ticks` on a downstream project, leading to a blank plot window.

```
Python 3.6.13 |Enthought, Inc. (x86_64)| (default, Apr 13 2021, 19:22:11) 
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.42.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from chaco.ticks import auto_ticks
>>> auto_ticks(657022.2328451854, 577935.6054180114, 657022.2328451854, 577935.6054180114, "auto", use_endpoints=False)
/Users/mdickinson/Enthought/ETS/chaco/chaco/ticks.py:392: RuntimeWarning: invalid value encountered in log10
  magnitudes = 10.0 ** floor(log10(candidate_intervals))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/mdickinson/Enthought/ETS/chaco/chaco/ticks.py", line 308, in auto_ticks
    ticks = arange(start, end + (tick_interval / 2.0), tick_interval)
ValueError: arange: cannot compute length
```

Expected behaviour: a list of tick locations is returned (possibly the empty list).

mdickinson:
Note that we have `data_low > data_high` for the inputs to `auto_ticks` above. While this may be indicative of an error in the calling code, I think it would make sense to also make `auto_ticks` robust for this kind of input.

corranwebster:
Not a fix, but a work-around: if you can create a replacement without the error, it is straightforward to create a new tick generator around the replacement and use that in the downstream app.

corranwebster:
Also, separately, this feels like it is a symptom, not the root cause: the axis bounds being inverted smells like there a problem with the values being passed to the ticking system.  In particular, are we in a situation where the width (or height, depending on orientation) of the axis component is negative?  If so, then all sorts of things may be messed up beyond ticking.

aaronayres35:
FWIW - 
This same traceback was seen a while back: 
https://github.com/enthought/chaco/issues/529
and that issue was closed by https://github.com/enthought/chaco/pull/636
but 636 was just checking for NaN inputs, _not non-NaN inputs that could lead to NaNs inside the method_

In this case, `tick_interval` on this line ends up as NaN
[...]

PR #848, opened 2023-01-19
Prevent axis related plot crash when plot is resized too small

aaronayres35:
This PR includes a defensive programming "fix" to #824 to prevent plot crashing.  The real fix would be to track down where the bad inputs are coming from.  However, regardless I think it makes sense for auto_ticks to be more robust to bad inputs as was discussed on the issue.

Also, in the same downstream project, when an application is resized to be very small and an inspector pane visibility is toggled, the plot some how ends up in a situation where `datalow > datahigh` in `chaco.axis.py` and we are hitting the `RunTimeError`.  I believe it makes sense to handle this case the same that we do for where `datalow == datahigh` (namely there is no plot we could show so we have no ticks).
I am unable to cause the downstream project's plots to crash when working off of this branch.

### DIFF
--- a/chaco/axis.py
+++ b/chaco/axis.py
@@ -500,7 +500,7 @@ class PlotAxis(AbstractOverlay):
             screenlow, screenhigh = screenhigh, screenlow
 
         if (
-            (datalow == datahigh)
+            (datalow >= datahigh)
             or (screenlow == screenhigh)
             or (datalow in [inf, -inf])
             or (datahigh in [inf, -inf])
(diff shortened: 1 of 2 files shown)


corranwebster:
Do we know which data range/mapper is giving the bad values?  I'm worried that there may be other places which are relying on the datalow <= datahigh invariant and so this may just push the error elsewhere.

In any case additional logging of the case datalow > datahigh (equal is OK) may be worthwhile.

corranwebster (approved):

PR #849, opened 2023-01-23
Fix CI on latest ubuntu

aaronayres35:


corranwebster (approved):
This is good for this phase - may need additional changes for Python 3.8/PySide 6 but that's OK

PR #850, opened 2023-01-23
Fix CI for Ubuntu latest version

corranwebster:
This aims to get CI back to a working state on current ubuntu.

### DIFF
--- a/.github/workflows/ets-from-source.yml
+++ b/.github/workflows/ets-from-source.yml
@@ -16,7 +16,7 @@ jobs:
   test-ets-source-linux:
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


aaronayres35:
Ah I'd opened #849 - but it looks like I was missing some changes added here.  I am happy to close my 849

corranwebster:
> Ah I'd opened #849 - but it looks like I was missing some changes added here. I am happy to close my 849

I'm not quite 100% sure what set of packages are needed - some of these may be what is needed to support Python3.8/PySide6, but that should be a different PR.

corranwebster:
Going to close this in favour of #849

Issue #846, opened 2022-12-13
ScatterPlot example currently shows marker_size takes integer arrays


mechworrior:
**Problem Description**
ScatterPlot example `scatter_1d` currently shows `marker_size` takes integer arrays as an input 
This will return a typeerror in the Kiva backend.  

**Reproduction Steps:**

Run the `scatter_1d` demo from etsdemo. with `marker_size=randint(1,5, numpts)`

**Expected behavior:**

This should have generated a scatterplot with different sized points.

**OS, Python version:** [windows, Python3.8]

mdickinson:
I can reproduce on macOS. If I change the line `        marker_size=3,  # randint(1,5, numpts),` (line 78 of the current demo script) to `        marker_size=randint(1,5, numpts),` and then run the demo script, I get a blank plot, and the following traceback on the console (repeated several times):

```python
Traceback (most recent call last):
  File "/Users/mdickinson/.edm/envs/testing/lib/python3.8/site-packages/enable/qt4/base_window.py", line 255, in paintEvent
    self.handler.paintEvent(event)
  File "/Users/mdickinson/.edm/envs/testing/lib/python3.8/site-packages/enable/qt4/base_window.py", line 90, in paintEvent
    self._enable_window._paint(event)
  File "/Users/mdickinson/.edm/envs/testing/lib/python3.8/site-packages/enable/abstract_window.py", line 536, in _paint
    self.component.draw(gc, view_bounds=(0, 0, size[0], size[1]))
  File "/Users/mdickinson/.edm/envs/testing/lib/python3.8/site-packages/enable/component.py", line 410, in draw
    self._draw(gc, view_bounds, mode)
  File "/Users/mdickinson/.edm/envs/testing/lib/python3.8/site-packages/enable/component.py", line 781, in _draw
    self._dispatch_draw(layer, bb, view_bounds, mode)
  File "/Users/mdickinson/.edm/envs/testing/lib/python3.8/site-packages/enable/container.py", line 270, in _dispatch_draw
    component._dispatch_draw(layer, gc, new_bounds, mode)
  File "/Users/mdickinson/.edm/envs/testing/lib/python3.8/site-packages/enable/component.py", line 809, in _dispatch_draw
    handler(gc, view_bounds, mode)
  File "/Users/mdickinson/.edm/envs/testing/lib/python3.8/site-packages/chaco/base_xy_plot.py", line 495, in _draw_plot
    self._render(gc, pts)
[...]

PR #859, opened 2023-03-13
MAINT: fix marker size type mismatch error

homosapien-lcy:
The ScatterPlot1D specifies marker_size = Float(4.0) while the downstream render_markers function is supposed to handle both single numeric data and arrays. This PR changes this.

Address #846


homosapien-lcy:
close issue #846

chaco/plots/scatterplot_1d.py
```diff
@@ -32,7 +32,9 @@ class ScatterPlot1D(Base1DPlot):
     marker = MarkerTrait
 
     # The pixel size of the marker, not including the thickness of the outline.
-    marker_size = Float(4.0)
+    marker_size = Union(Float, Int,
+                        Array(Float), Array(Int),
+                        default_value=None)
```
dpinte, line 37: Why changing the defaut to None? The 4.0 value was a good default. By changing the default, you now have to update the `_render` method.

The upstream issue was related to the fact that only floats were supported. Using a Union type, you now allow other data type, which is solving the problem. Adding a tests that reproduces the uptsream problem would be good

dpinte (changes requested):

chaco/plots/scatterplot_1d.py
homosapien-lcy: I somehow remembered that using 4.0 as default will raise an error, but after retest, works fine... Will use 4.0 instead.

chaco/plots/scatterplot_1d.py
```diff
@@ -32,7 +32,9 @@ class ScatterPlot1D(Base1DPlot):
     marker = MarkerTrait
 
     # The pixel size of the marker, not including the thickness of the outline.
-    marker_size = Float(4.0)
+    marker_size = Union(Float, Int,
```
mdickinson, line 35: The `Float` trait should already accept integers, so I'm not sure that it adds much to accept `Int` here, especially given that the marker size is conceptually a float here.

chaco/plots/scatterplot_1d.py
```diff
@@ -32,7 +32,9 @@ class ScatterPlot1D(Base1DPlot):
...
+                        Array(Float), Array(Int),
```
mdickinson, line 36: This isn't how the `Array` trait works - you need to supply a dtype as the argument, not a trait type. As it stands, this would accept an array with any dtype.

chaco/plots/scatterplot_1d.py
dpinte: @homosapien-lcy adding a new manual example is not adding a test! 
As discussed, tests are based on the `unittest.TestCase` in order too be picked up by the CI and shoud be added here: https://github.com/enthought/chaco/blob/main/chaco/plots/tests/test_scatterplot_1d.py

dpinte:
@homosapien-lcy as next step, we discussed:
- [ ] add a test case reproducing the upstream issue, that should surface the problem that @mdickinson has highlighted with the way you use the `Array` trait
- [ ] fix the current version of the code and make sure the test pass
- [ ] add tests cases for the various other use case you cover in the new `Union` (and verify that as Mark suggests, it's not needed to add the `Int` trait in the `Union`)

chaco/plots/scatterplot_1d.py
homosapien-lcy: Yeah, I found that Float trait can accomodate int as well, just cahnged.

chaco/plots/scatterplot_1d.py
homosapien-lcy: Got it, thanks!

chaco/tests/test_scatter_1d.py
```diff
@@ -0,0 +1,139 @@
...
+"""
+
+from numpy import sort
+from numpy.random import random, randint, uniform
+import unittest
+from chaco.api import ArrayPlotData, Plot
+
+
+class Scatter1DTestCase(unittest.TestCase):
+    # test the default value 4.0
```
dpinte, line 22: that comment line can be removed in favour of a clear method name or moved as a docstring to the test method.

chaco/tests/test_scatter_1d.py
```diff
@@ -0,0 +1,139 @@
...
+    def test_default(self):
```
dpinte, line 23: I would suggests `test_default_marker_size` rather than `test_default`

chaco/tests/test_scatter_1d.py
```diff
@@ -0,0 +1,139 @@
...
+        plot.plot_1d(
+            "value",
+            type="scatter_1d",
+            orientation="v",
+            marker="plus",
+            alignment="left"
+        )
+
+    # test integer marker size
+    def test_int(self):
```
dpinte, line 46: same comment on the method name and the removal of the comment

chaco/tests/test_scatter_1d.py
```diff
@@ -0,0 +1,139 @@
...
+        # Create the plot
+        plot = Plot(pd, use_backbuffer=True, auto_grid=False)
+
+        plot.plot_1d(
+            "value",
+            type="scatter_1d",
+            orientation="v",
+            marker="plus",
+            alignment="left"
+        )
```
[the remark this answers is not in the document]
dpinte: that is applicable to all the tests.

dpinte (changes requested):
@homosapien-lcy some comments to address on the test.

chaco/tests/test_scatter_1d.py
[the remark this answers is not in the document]
homosapien-lcy: Yes, these tests are limited to check for traits mismatch exceptions. Is there other stuffs we want to test?

chaco/tests/test_scatter_1d.py
homosapien-lcy: Thanks for the comments, updates made

chaco/plots/scatterplot_1d.py
```diff
@@ -32,7 +32,9 @@ class ScatterPlot1D(Base1DPlot):
     marker = MarkerTrait
 
     # The pixel size of the marker, not including the thickness of the outline.
-    marker_size = Float(4.0)
+    marker_size = Union(Float,
+                        Array(dtype=float),
+                        default_value=4.0)
```
corranwebster, line 37: This is absolutely not correct - Chaco data never comes from hard-coded arrays like this.  Most likely this should be a Float or an `Instance(AbstractDataSource)`.

Also make sure that you keep the default value of 4.0 otherwise lots of plots will break.

chaco/examples/demo/basic/scatter_1d.py
```diff
@@ -69,6 +69,15 @@ def _create_plot_component():
         alignment="left",
     )
 
+    plot.plot_1d(
+        "value",
+        type="scatter_1d",
+        orientation="v",
+        marker="plus",
+        alignment="left",
+        marker_size=randint(1,5, numpts)
```
corranwebster, line 78: This is not the correct usage and should fail.  You would need to specify a data source for the size values which holds the sizes.

chaco/tests/test_scatter_1d.py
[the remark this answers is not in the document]
corranwebster: If this is a test of the default marker size, you should somewhere have a line like `self.assertEqual(..., 4.0)`

With the code as you have it now, it would fail!

chaco/tests/test_scatter_1d.py
```diff
@@ -0,0 +1,134 @@
...
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+
+"""
+Unit tests for scatter_1d function
```
corranwebster, line 12: There are already tests for the 1D scatterplot here: https://github.com/enthought/chaco/blob/main/chaco/plots/tests/test_scatterplot_1d.py

corranwebster (changes requested):
Not really sure what the point of this PR is.  The 1D scatterplot doesn't currently support variable size markers (or variable color or shape, for that matter).  It could, potentially, but it doesn't right now, and this PR is not correct in its approach.  Chaco is a dynamoc plotting library and all data needs to be passed via appropriate objects that manage changes to the data.  For example, the example usage will fail if the using code changes the _size_ of the data array holding the values.

To be fair, it looks like there is an error in the corresponding Trait declarations for the 2D scatterplot: the following is not correct either (edit: as is noted in the "TODO") https://github.com/enthought/chaco/blob/f32e01cd543bb4b7ca144808f578df5c768cd833/chaco/plots/scatterplot.py#L228-L231

A correct implementation would use a data source for the sizes (and possibly a mapper as well), the way that the colormapped scatterplot does for colors:
https://github.com/enthought/chaco/blob/f32e01cd543bb4b7ca144808f578df5c768cd833/chaco/plots/colormapped_scatterplot.py#L65-L66

It would also have to add support for finding the data source in the `Plot` object code, eg. somewhere in here: https://github.com/enthought/chaco/blob/f32e01cd543bb4b7ca144808f578df5c768cd833/chaco/plot.py#L1270-L1296

homosapien-lcy:
With Corran's comments, after discussing with Didrik, we think:

1 the "error" that the user reported in #846 is actually the expected behavior of the test, thus there is no need to change the scatterplot class to enable different types of inputs
2 the confusion comes from the fact that in chaco/examples/demo/basic/scatter_1d, a line "# randint(1,5, numpts)," is commented out in line 78, suggesting the users that the input may be an array of integers.

With the above observations, we decided to simply delete the commented out line mentioned in 2 to avoid confusion as the solution to the issue closes #846

corranwebster (approved):

dpinte (approved):

PR #864, opened 2023-03-30
fix world_map demo with new link

homosapien-lcy:
The chaco/examples/demo/world_map demo example is broken due to the original image link in the example has expired. This PR replaces it with a new example.

### DIFF
--- a/examples/demo/world_map.py
+++ b/examples/demo/world_map.py
@@ -8,10 +8,6 @@ negative values.
    region to zoom.  If you use a sequence of zoom boxes, pressing alt-left-arrow
    and alt-right-arrow moves you forwards and backwards through the "zoom
    history".
-
-The image in this demo is an image from wikipedia
-https://commons.wikimedia.org/wiki/File:Winkel_triple_projection_SW.jpg
-Copyright Attribution: Strebe, Public domain, via Wikimedia Commons
 """
 
 # Standard library imports
--- a/examples/demo/world_map.py
+++ b/examples/demo/world_map.py
@@ -8,6 +8,10 @@ negative values.
    region to zoom.  If you use a sequence of zoom boxes, pressing alt-left-arrow
    and alt-right-arrow moves you forwards and backwards through the "zoom
    history".
+
+The image in this demo is an image from wikipedia
+https://commons.wikimedia.org/wiki/File:Winkel_triple_projection_SW.jpg
+Copyright Attribution: Strebe, Public domain, via Wikimedia Commons
 """
 
 # Standard library imports
--- a/examples/demo/world_map.py
+++ b/examples/demo/world_map.py
@@ -38,7 +38,7 @@ class WorldMapPlot(HasTraits):
 
     # The URL which points to the world map image to be downloaded
     image_url = Str(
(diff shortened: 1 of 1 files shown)


dpinte:
Can you add a comment to the example with the attribution of the image copyright. From wikimedia, it sounds like "Strebe, Public domain, via Wikimedia Commons" would be good and maybe a link to the page (https://commons.wikimedia.org/wiki/File:Winkel_triple_projection_SW.jpg)

examples/demo/world_map.py
```diff
@@ -34,7 +38,7 @@ class WorldMapPlot(HasTraits):
 
     # The URL which points to the world map image to be downloaded
     image_url = Str(
-        "http://eoimages.gsfc.nasa.gov/ve//2433/land_shallow_topo_2048.jpg"
+        "https://upload.wikimedia.org/wikipedia/commons/9/91/Winkel_triple_projection_SW.jpg"
```
corranwebster, line 41: This uses a different projection, and so the plotted points are not correct (eg. Austin will not be plotted correctly).

The original image is this one in Wikimedia: https://commons.wikimedia.org/wiki/File:Land_shallow_topo_2048.jpg

So this URL may be the right one to use (although it is very high resolution): https://upload.wikimedia.org/wikipedia/commons/9/91/Land_shallow_topo_2048.jpg

corranwebster: And this is the source from NASA: https://visibleearth.nasa.gov/images/57752/blue-marble-land-surface-shallow-water-and-shaded-topography/57754l

So this may be even better: https://eoimages.gsfc.nasa.gov/images/imagerecords/57000/57752/land_shallow_topo_2048.jpg

corranwebster (changes requested):

examples/demo/world_map.py
homosapien-lcy: Thanks, just changed to "https://eoimages.gsfc.nasa.gov/images/imagerecords/57000/57752/land_shallow_topo_2048.jpg" as you proposed. I assume we don't need a copyright header since it's the same NASA image as before?

dpinte (approved):
LGTM

homosapien-lcy:
So, is this issue ready for close?

corranwebster (approved):
Looks good to me!

corranwebster:
You might want to merge in main to check that everything passes.

homosapien-lcy:
Somehow I still cannot merge this due to CI test error (which comes from MacOS RuntimeError: NumPy was built with baseline optimizations: 
(SSE SSE2 SSE3 SSSE3 SSE41 POPCNT SSE42 AVX F16C AVX2) but your machine doesn't support:
(AVX2).), I think this is an issue we can pass for now?

corranwebster:
> Somehow I still cannot merge this due to CI test error (which comes from MacOS RuntimeError: NumPy was built with baseline optimizations:
> (SSE SSE2 SSE3 SSSE3 SSE41 POPCNT SSE42 AVX F16C AVX2) but your machine doesn't support:
> (AVX2).), I think this is an issue we can pass for now?

Yes, this is due to a change in the EDM mac numpy build.  We need to turn off some of the mac tests.

corranwebster:
I've made the MacOS tests no longer required, so this should be able to be merged despite the failure.

PR #870, opened 2023-03-31
MAINT: replace deprecated boston dataset with diabetes

homosapien-lcy:
The load_boston dataset is deprecated, replacing with the diabetes dataset (seems to the the only usable dataset in sklearn.dataset, iris and california housing dataset is not compatible with the code).

**However, there are two more errors arised after the fix that I found the fixes but want to discuss:**

**After correcting the dataset, an error will happen in get_image_from_file() since this function tries to read an image in the demo/basic/capitol.jpg, but the folder doesn't seem to exist anymore, removing this function can fix this problem.**

**Then, another error will happend in chaco.chaco.plots.polar_line_renderer, in function _draw_plot(). Function will try to call _render on gc (line 136), which is a parameter not exists. Removing this line can solve the problem** (this also causes exception in issue #875 )

**When both the above fixes are done, the plots can be generated without error. But are these fixes sounds good to you?**

### DIFF
--- a/examples/user_guide/plot_types/create_plot_snapshots.py
+++ b/examples/user_guide/plot_types/create_plot_snapshots.py
@@ -149,9 +149,9 @@ get_line_plot_connectedhold = partial(get_line_plot, "connectedhold")
 
 
 def get_scatter_plot():
-    boston = datasets.load_boston()
-    prices = boston["target"]
-    lower_status = boston["data"][:, -1]
+    diabetes = datasets.load_diabetes()
+    prices = diabetes["target"]
+    lower_status = diabetes["data"][:, -1]
 
     x, y = get_data_sources(x=lower_status, y=prices)
     x_mapper, y_mapper = get_mappers(x, y)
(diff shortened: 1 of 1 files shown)


dpinte:
The capitol.jpg image is still available here: https://github.com/enthought/chaco/blob/cbf53e733e5cbfae6338b46faf29032cf73bf899/chaco/examples/demo/basic/capitol.jpg

dpinte:
@homosapien-lcy you should update the code to use the importlib.resources to access the file.

homosapien-lcy:
Will use importlib_resource to find capitol.jpg

dpinte:
@homosapien-lcy by the way, I think it's important to split the issues into different PR's. Everything related to fixing the create_plot_snapshot can be done in this PR (but then the PR must be renamed). The issue with polar renderer is not a problem with the demo but a bug in chaco. That should be a different PR!

homosapien-lcy:
> @homosapien-lcy by the way, I think it's important to split the issues into different PR's. Everything related to fixing the create_plot_snapshot can be done in this PR (but then the PR must be renamed). The issue with polar renderer is not a problem with the demo but a bug in chaco. That should be a different PR!

Got it, separating into two other PRs

Issue #876, opened 2023-03-31
ImportError: attempted relative import with no known parent package


homosapien-lcy:
**Problem Description**
ImportError: attempted relative import with no known parent package when running chaco/examples/demo/zoomed_plot/zoom_plot.py

**Reproduction Steps:**
python chaco/examples/demo/zoomed_plot/zoom_plot.py

**OS, Python version:** [MacOS, python3.8 (with EDM), python3.11]

PR #877, opened 2023-03-31
Fix parental import

homosapien-lcy:
fix the parental import error. Closes issue #876

### DIFF
--- a/examples/demo/zoomed_plot/zoom_plot.py
+++ b/examples/demo/zoomed_plot/zoom_plot.py
@@ -24,7 +24,7 @@ from chaco.api import VPlotContainer, ArrayPlotData, Plot, PlotGrid, PlotAxis
 from chaco.tools.api import RangeSelection
 
 # Relative imports
-from .zoom_overlay import ZoomOverlay
+from zoom_overlay import ZoomOverlay
 
 sample_path = os.path.join("examples", "data", "sample.wav")
 alt_path = os.path.join("..", "data", "sample.wav")


examples/demo/zoomed_plot/zoom_plot.py
```diff
@@ -24,7 +24,7 @@
 from chaco.tools.api import RangeSelection
 
 # Relative imports
-from .zoom_overlay import ZoomOverlay
+from zoom_overlay import ZoomOverlay
```
dpinte, line 27: looking at the folder structure, zoom_overlay is part of the demo.zoomed_plot package. It would be better to make it an absolute import.

dpinte (changes requested):

examples/demo/zoomed_plot/zoom_plot.py
homosapien-lcy: I tried changing it to an absolute import:
from examples.demo.zoomed_plot.zoom_overlay import ZoomOverlay

However that will yield "ModuleNotFoundError: No module named 'examples'", even though the vscode shows the import path is correct... Do you know why it does that?

corranwebster: `examples` isn't a standard Python package (no `__init__.py`) so it will only be picked up as a module when it a subdirectory of something in `sys.path`.  That's good, we don't _want_ Chaco to have a top-level `examples` package.  But VSCode is probably running its analysis with `sys.path` including the top-level repo directory, so it picks up `examples` (technically examples is a "namespace package" in this case - gory details are in https://peps.python.org/pep-0420/)

homosapien-lcy: I tried the 4 below:

1. zoom_overlay
2. zoomed_plot.zoom_overlay
3. demo.zoomed_plot.zoom_overlay
4. examples.demo.zoomed_plot.zoom_overlay

Only the first one (relative import) work, should we just go with that? @dpinte

dpinte: Running from the `examples` directory, the following fails:
```
$ python -m demo.zoomed_plot.zoom_plot
/Users/dpinte/.edm/envs/edm38/lib/python3.8/site-packages/chaco/color_mapper.py:347: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if self.color_depth is "rgba":
/Users/dpinte/.edm/envs/edm38/lib/python3.8/site-packages/chaco/tools/toolbars/plot_toolbar.py:238: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if self.location is "top":
/Users/dpinte/.edm/envs/edm38/lib/python3.8/site-packages/chaco/tools/toolbars/plot_toolbar.py:251: SyntaxWarning: "is" with a literal. Did you mean "=="?
  elif self.location is "bottom":
/Users/dpinte/.edm/envs/edm38/lib/python3.8/site-packages/chaco/tools/toolbars/plot_toolbar.py:258: SyntaxWarning: "is" with a literal. Did you mean "=="?
  elif self.location is "left":
  File "/Users/dpinte/.edm/envs/edm38/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/Users/dpinte/.edm/envs/edm38/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/Users/dpinte/projects/enthought/chaco/examples/demo/zoomed_plot/zoom_plot.py", line 27, in <module>
    from zoom_overlay import ZoomOverlay
ModuleNotFoundError: No module named 'zoom_overlay'
```

Issue #867, opened 2023-03-30
ValueError: 2D bounds of an image plot needs to have the same shape as the underlying data, because they are assumed to be generated from meshgrids. when running chaco/examples/tutorials/scipy2008/traits_image.py


homosapien-lcy:
**Problem Description**
ValueError: 2D bounds of an image plot needs to have the same shape as the underlying data, because they are assumed to be generated from meshgrids. when running chaco/examples/tutorials/scipy2008/traits_image

**Reproduction Steps:**

```python
python chaco/examples/tutorials/scipy2008/traits_image.py
```

**Expected behavior:**
pass test

**OS, Python version:** [MacOS, python3.11 or python3.8 (with or without EDM)]

homosapien-lcy:
The error happens in line 898, the upper stream set cell_plot=True, cause num_ticks = array_data.shape[axis] + 1 which is one off from bounds.shape[axis]

homosapien-lcy:
This seemed to be caused by traits_image specifying xbound to be of size 50 upstream which is in conflict with the num_ticks

PR #882, opened 2023-04-04
Image size fix

homosapien-lcy:
closes issue #867 
The demo chaco/examples/tutorials/scipy2008/traits_image.py is broken as mentioned in the issue.
Since this is a very early example (2008), it's likely some updates in chaco has broken this demo. I found that all recent calls to img_plot (line 678, chaco/chaco/plot.py) uses a range other than mash for xbounds and ybounds since plot._process_2d_bounds will regenerate the mash with the range that has the correct dimension. Thus I change this example to use a range other than a mesh as well.

### DIFF
--- a/examples/tutorials/scipy2008/traits_image.py
+++ b/examples/tutorials/scipy2008/traits_image.py
@@ -38,8 +38,8 @@ class ImagePlotTraits(HasTraits):
         self.renderer = plot.img_plot(
             "imagedata",
             name="plot1",
-            xbounds=xgrid,
-            ybounds=ygrid,
+            xbounds=(0, 6),
+            ybounds=(0, 8),
             colormap=viridis,
         )[0]
         self.plot = plot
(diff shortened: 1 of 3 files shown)


examples/tutorials/scipy2008/traits_image.py
```diff
@@ -38,8 +38,8 @@ def __init__(self):
         self.renderer = plot.img_plot(
             "imagedata",
             name="plot1",
-            xbounds=xgrid,
-            ybounds=ygrid,
+            xbounds=(0, 6),
```
mdickinson, line 41: Why are the bounds given as `0` and `6` when `x` goes from`0` to `8`?

homosapien-lcy: Yeah, it should be the other way around... Just fixed

dpinte (approved):

corranwebster (approved):

PR #883, opened 2023-04-04
fix gc undeclare

### LINKED ISSUE #875: NameError: name 'gc' is not defined
**Problem Description**
NameError: name 'gc' is not defined in chaco/chaco/plots/polar_line_renderer.py

**Reproduction Steps:**
python chaco/examples/demo/simple_polar.py

[Steps to reproduce issue here.]

```python
# Code to reproduce issue here
```

**Expected behavior:**  

[MacOS, python3.8 (EDM)]

homosapien-lcy:
Found that the gc (GraphicsContext) is referenced before construction in the line 136. This has caused the test chaco/chaco/examples/demo/simple_polar.py to fail.
The solution is to find the gc in the context. A gc context is actually conveniently included in the *args, thus retrieving that gc as GraphicContext to solve the problem. closes #875

### DIFF
--- a/chaco/plots/polar_line_renderer.py
+++ b/chaco/plots/polar_line_renderer.py
@@ -130,10 +130,9 @@ class PolarLineRenderer(AbstractPlotRenderer):
     def _downsample(self):
         return self.map_screen(self._cached_data_pts)
 
-    def _draw_plot(self, *args, **kw):
+    def _draw_plot(self, gc, *args, **kw):
         """Draws the 'plot' layer."""
         self._gather_points()
-        gc = args[0]
         self._render(gc, self._cached_data_pts)
 
     def _bounds_changed(self, old, new):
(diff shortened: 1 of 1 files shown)


chaco/plots/polar_line_renderer.py
```diff
@@ -133,6 +133,7 @@ def _downsample(self):
     def _draw_plot(self, *args, **kw):
         """Draws the 'plot' layer."""
         self._gather_points()
+        gc = args[0]
```
corranwebster, line 136: If you want the `gc` to be the first argument to the function, make it the first argument to the function!

Doing this is fragile, not self-documenting and will produce confusing error messages if used incorrectly.

corranwebster (changes requested):
Just add `gc` as the first argument to the function

corranwebster (approved):

dpinte:
For reference, the prototype of the `_draw_plot` method is defined here: https://github.com/enthought/enable/blob/main/enable/component.py#L807.  You could even remove the *arg and **kwargs in favour of the ones from the prototype.

Issue #888, opened 2023-05-02
traits.trait_errors.TraitError: Cannot set the undefined 'format' attribute of a 'RangeEditor' object. when running chaco/chaco/examples/demo/nonlinear_color_mapping.py


homosapien-lcy:
**Problem Description**
When running example chaco/chaco/examples/demo/nonlinear_color_mapping.py, a traits.trait_errors.TraitError: Cannot set the undefined 'format' attribute of a 'RangeEditor' object. is raised

**Reproduction Steps:**
```python
python chaco/chaco/examples/demo/nonlinear_color_mapping.py
```

**Link:** 
https://github.com/enthought/chaco/blob/main/chaco/examples/demo/nonlinear_color_mapping.py

**Full Stacktrace:**
```
(py311) (base) cyliu@aus552cyliu Documents % python3.11 3.11_test/chaco/chaco/examples/demo/qt_example.py
2023-05-02 17:34:47.993 Python[5398:120343] ApplePersistenceIgnoreState: Existing state will not be touched. New state will be written to /var/folders/2z/kylzj9s92y71cxscmljmpqrh0000gt/T/org.python.python.savedState
(py311) (base) cyliu@aus552cyliu Documents % python3.11 3.11_test/chaco/chaco/examples/demo/nonlinear_color_mapping.py 
Traceback (most recent call last):
  File "/Users/cyliu/Documents/3.11_test/chaco/chaco/examples/demo/nonlinear_color_mapping.py", line 169, in <module>
    class DataGridView(HasTraits):
  File "/Users/cyliu/Documents/3.11_test/chaco/chaco/examples/demo/nonlinear_color_mapping.py", line 198, in DataGridView
    editor=RangeEditor(low=0.1, high=3.0, format="%4.2f"),
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/cyliu/Documents/3.11_test/traitsui/traitsui/editor_factory.py", line 81, in __init__
    HasPrivateTraits.__init__(self, **traits)
traits.trait_errors.TraitError: Cannot set the undefined 'format' attribute of a 'RangeEditor' object.
```

**OS, Python version:** [MacOS, python3.11]

Package            Version     Editable project location
------------------ ----------- -----------------------------------------
apptools           5.2.0
attrdict           2.0.1
attrdict3          2.0.2
blosc2             2.0.0
celiagg            2.1.3
certifi            2022.12.7
[...]

PR #892, opened 2023-05-08
MAINT: fix depreciated formt usage in nonlinear_color_mapping

homosapien-lcy:
After the update of the traitsui package, RangeEditor has deprecated the usage of format_str. Some code that didn't adapt to the change has raised error during the test such as mentioned in #888 . 
The error is located in https://github.com/enthought/chaco/blob/main/chaco/examples/demo/nonlinear_color_mapping.py#L198, and the current PR fixes it by updating the deprecated "format" to "format_str".


dpinte (approved):
LGTM

PR #894, opened 2023-05-08
MAINT: fix the function interpretation for scalar_image_function_inspector.py

homosapien-lcy:
After the update of python version, the how "exec" works seem have changed in certain versions and caused exec("import *") to fail. This has caused issue in issue #890 . To address this issue, I changed the import to include specific name of the modules (cos, tanh, jn) in use. Closes issue #890


corranwebster (changes requested):
No, the difference is that last year `scipy.__init__` has added an `__all__` to the module, which restricts what is imported by  `from scipy import *` so that it no longer includes the basic functions like `tanh`.

Given that the formula is meant to be edited by humans and to allow them to enter any reasonable scientific function, we want to keep the "*" imports, but we should probably get basic functions from NumPy and special functions from SciPy.

corranwebster (approved):

PR #895, opened 2023-05-10
Handle `nan` in barplots.

### LINKED ISSUE #845: Bar plot disappears and "Exceeded cell block limit" traceback when encountering NaN
**Problem Description**

When plotting (`type="bar"`) data with NaN, the plot goes blank and a traceback is emitted at paint-time.

**Reproduction Steps:**

```python
import numpy
from traits.api import *
from traitsui.api import *
from chaco.api import *
from chaco.tools.api import *

class PlotExample(HasTraits):
    plot = Instance(Plot)

    def _plot_default(self):
        index = numpy.arange(10.)
        series = index * 10
        series[-1] = float("nan")
        plot_data = ArrayPlotData(index=index)
        plot_data.set_data("series", series)
        plot = Plot(plot_data)
        plot.plot(("index", "series"), type="bar", bar_width=0.8)

        # optional: NaN is initially out of the visible range, so it only crashes upon panning
        plot.index_range.low_setting = -0.5
        plot.index_range.high_setting = 6.5
        plot.tools.append(PanTool(plot, constrain=True, constrain_direction="x"))
        return plot

    traits_view = View(
        UItem("plot", editor=ComponentEditor()),
        width=400, height=400, resizable=True,
    )

demo = PlotExample()
demo.configure_traits()
```

chaco_crash

**Expected behavior:**

(For instance) Emit a warning when NaNs are encountered in bar plot data, possibly not at paint-time but as soon as the plot data is updated with offending values. Then avoid plotting the corresponding bar. 

Python 3.8, OS irrelevant, backend `agg`.

corranwebster:
Also somewhat better handling of cropping rectangles that are offscreen for common cases.

Includes basic tests.

Fixes #845

### DIFF
--- a/chaco/plots/barplot.py
+++ b/chaco/plots/barplot.py
@@ -18,7 +18,7 @@ from numpy import (
     column_stack,
     empty,
     invert,
-    isnan,
+    isfinite,
     transpose,
     zeros,
 )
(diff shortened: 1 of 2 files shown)


dpinte (approved):
LGTM

PR #900, opened 2023-05-12
Fire Range `updated` event when data sources are set to empty

### LINKED ISSUE #821: Removing a data source makes `DataView` objects inconsistent
**Problem Description**
Removing a data source from a `DataView` instance leaves it internally inconsistent.

**Reproduction Steps:**
This snippet demonstrates the inconsistency:
```python
import numpy as np

from chaco.api import DataView
from chaco.grid_data_source import GridDataSource

source = GridDataSource(xdata=np.array([1, 2, 4]), ydata=np.array([7, 8, 9]))
component = DataView()
print(f"Initial output: {component.map_screen(np.array([0, 1]))}")
component.range2d.add(source)  # (1)
component.map_screen(np.array([0, 1]))  # (2)
component.range2d.remove(source)  # (3)
print(f"Final output: {component.map_screen(np.array([0, 1]))}")
```
Output:
```
Initial output: [[0. 0.]]
Final output: [inf inf]
```

Initially, without any data sources, the output of `component.map_screen` is `[[0, 0]]`. When a data source is added and then removed, the bounds on `component.value_mapper.range` to refreshed to (-inf, inf). However, `component.value_mapper._cache_valid` isn't reverted from `True` to `False`. (It's set to `True` as a consequence of (2) in the snippet above.) As a result, subsequent calls to `component.map_screen` produce `inf`s. This `inf`s in turn have the potential to turn into `nan`s, raising exceptions in unexpected places downstream.

**Expected behavior:**
I *think* the output should be [[0., 0.]] again, (or more precisely, `component.value_mapper._cache_valid` should revert to `False` if `component.value_mapper.range.refresh()` is called.  However,  I'm not sure and opening this issue for discussion.

**OS, Python version:** macOS Catalina 10.15.7, Python 3.6.13

corranwebster:
The `updated` event on a `DataRange1D` was not being fired when the last data source was removed.  This fixes that.

Fixes #821

### DIFF
--- a/chaco/data_range_1d.py
+++ b/chaco/data_range_1d.py
@@ -352,6 +352,7 @@ class DataRange1D(BaseDataRange):
                 self._high_value = inf
             else:
                 self._high_value = self._high_setting
+            self.updated = (self._low_value, self._high_value)
             return
         else:
             mins, maxes = zip(*bounds_list)
--- a/chaco/tests/test_datarange_1d.py
+++ b/chaco/tests/test_datarange_1d.py
@@ -15,6 +15,7 @@ from numpy import arange, array, zeros, inf
 from numpy.testing import assert_equal
 
 from traits.api import HasTraits, Instance, Bool, observe
+from traits.testing.api import UnittestTools
 
 from chaco.api import DataRange1D, ArrayDataSource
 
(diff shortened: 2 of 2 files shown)


dpinte (approved):
:+1:

PR #903, opened 2023-05-15
Remove deprecated stubs

### LINKED ISSUE #748: Remove deprecated stub modules in chaco 6.0
ref: #738 and #741

In the recent reorg, old plot and overlay modules were left at their old locations as stubs to preserve backwards compatibility.  These can be removed in the next major release.

corranwebster:
This removes plot and overlay stub modules as promised in their deprecation warnings.

Fixes #748


dpinte (approved):
:+1:

PR #905, opened 2023-05-15
Update TextPlot when color, font or other aesthetics change.

### LINKED ISSUE #729: changing TextPlot color & font
Hello,

I would like to change TextPlot color & font.

Initialization is ok, but when the color trait is changed for instance, the color is not changed in the display.

There is the same issue with the font trait.

Here's the CME.

```python
#! /usr/bin/env python3

import numpy as np

from traits.api import (
    Array,
    Color,
    Font,
    HasTraits,
    Instance,
    List
)
from traitsui.api import Item, UItem, View


class Data(HasTraits):

    data = Array
    labels = List
    color = Color
    font = Font
    plot = Instance(Plot)

    def _data_default(self):
        data = np.array([[0, 0], [0, 1], [1, 1], [1, 0]])
        return (data)

    def _labels_default(self):
        labels = ['A', 'B', 'C', 'D']
        return (labels)

    def _plot_default(self):
        self.plotdata = ArrayPlotData()
        self.plotdata.set_data('x', self.data[:, 0])
        self.plotdata.set_data('y', self.data[:, 1])
        self.plotdata.set_data('labels', self.labels)

        plot = Plot(self.plotdata)
        plot.range2d.set_bounds((-1, -1), (2, 2))
        plot.plot(("x", "y"),
                  type='scatter',
                  marker='dot',
                  color=self.color)
        plot.plot(("x", "y", "labels"),
                  type='text',
                  text_margin=4,
                  h_position='right',
                  text_offset=(4, 4),
                  text_color=self.color)

        return plot

    traits_view = View(
        UItem(
            "plot",
            editor=ComponentEditor(),
            resizable=True
        ),
        UItem('color',
              style='simple'),
        UItem('font'),
        resizable=True,
        buttons=["OK"],
        width=900,
        height=800,
    )

    def _color_changed(self):
        self.plot.plots['plot0'][0].color = self.color
        self.plot.plots['plot1'][0].text_color = self.color

    def _font_changed(self):
        name = self.font.family()
        size = self.font.pointSize()
        self.plot.plots['plot1'][0].text_font = '%s %d' % (name, size)


if __name__ == '__main__':
    viewer = Data(color=(255, 128, 64))
    viewer.configure_traits()
```

Thanks in advance for any help.

Regards

Debian 10 x86_64, Python 3.7.3, ETS source code from git repo

corranwebster:
Forces recreation of all the `Label` objects, which is suboptimal, but since this uses `Label` instances this is not a highly performant plot in any case.

Fixes #729


dpinte (approved):
:+1:

fred4ets:
Thanks a lot!

